PR #61, opened 2013-03-16
Many specific exceptions instead of one LdapError

DenisKnauf:
We need to rescue some exceptions, but because there is only one LdapError-class, we cannot rescue one specific behavior.
So i wrote this patch for using many LdapError-classes but specific exceptions.
All of these exceptions has LdapError as base-class, so it do not break any exception handling.

### DIFF
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -244,6 +244,33 @@ class Net::LDAP
   VERSION = "0.4.0"
 
   class LdapError < StandardError; end
+  class AlreadyOpenedError < LdapError; end
+  class SocketError < LdapError; end
+  class ConnectionRefusedError < LdapError; end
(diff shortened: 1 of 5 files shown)


satoryu:
I and my colleagues have been looking forward to this changes on LdapError :+1:  
Unfortunately @DenisKnauf seems be not able to have a time to fix the conflict. 

I have an idea to take over his work, creating new pull request base on a branch into which @DenisKnauf 's master are merged with the current HEAD of this repo's master, of course resolving the conflict. 

@mtodd @jch how do you think the idea?

mtodd:
@satoryu like @DenisKnauf, I too have not had time to fix this. I'd *love* for you to take this over and get it to a point where we can merge it! :+1:

jch:
Closing in favor of #183

PR #164, opened 2014-11-30
Non-blocking Socket Connect

mtodd:
This replaces the existing, blocking connection implementation with non-blocking connection logic.

More (tests) to come.

cc @jch @schaary


lib/net/ldap/connection.rb
```diff
@@ -28,6 +30,54 @@ def initialize(server)
...
+      if IO.select(nil, [socket], nil, timeout)
+        # validate connection succeeded
+        begin
+          socket.connect_nonblock(sockaddr)
+        rescue Errno::EISCONN
+          # already connected, no action necessary
+        rescue Exception => e
+          # unexpected error
+          socket.close
+          raise Net::LDAP::LdapError, "Connection to #{host}:#{port} (#{ip}) failed: (#{e.class}) #{e.message}"
```
satoryu: I think new exception class inheriting `Net::LDAP::Error` should be defined for this error.

HarlemSquirrel:
If I'm not mistaken, #273 added non-blocking connections when timeout is set so I think we no longer need this.

PR #166, opened 2014-12-01
Specify certfile to verify SSL certs against in tests

mtodd:
#162 added a TLS Start test but didn't assert verified certs because the `tls_options` hadn't been wired up yet. #161 changed this, but hadn't been tested against the additional test in #162 before getting merged, resulting in `master` failing.

It's easy enough to disable certificate verification, but it's not necessary. Instead, we can also set the `:ca_file` TLS options to verify against.

cc @jch @schaary @sonOfRa @tarcieri

### DIFF
--- a/test/integration/test_bind.rb
+++ b/test/integration/test_bind.rb
@@ -20,8 +20,14 @@ class TestBindIntegration < LDAPIntegrationTestCase
     refute @ldap.bind(method: :simple, username: "uid=user1,ou=People,dc=rubyldap,dc=com", password: "not my password"), @ldap.get_operation_result.inspect
   end
 
-  def test_bind_tls
+  def test_bind_tls_with_cafile
     tls_options = OpenSSL::SSL::SSLContext::DEFAULT_PARAMS.merge(:ca_file => CA_FILE)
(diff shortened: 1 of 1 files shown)


zmajstor:
there is a failing test:
```
Error: test_bind_tls_with_cafile(TestBindIntegration)
: OpenSSL::SSL::SSLError: SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed
```
I think that the reason is that [install-openldap] (https://github.com/ruby-ldap/ruby-net-ldap/blob/master/script/install-openldap#L59-L63) creates self-signed CA certificate on VM setup, so 
in order to make it pass, content of `fixtures/cacert.pem` have to be replaced with this one:
```
$ vagrant ssh

vagrant@rubyldap:~$ cat /etc/ssl/certs/cacert.pem
```

I guess there is a more elegant way to handle this ... but this works as a quick fix

tarcieri:
@zmajstor see my original notes on the PR: certificates should be generated for tests, not checked in.

I'm a big fan of the `certificate_authority` gem for this. Otherwise, there's `r509`, but it's MRI-specific which kinda sucks.

PR #167, opened 2014-12-01
Non-blocking reads with configurable timeouts

mtodd:
Rewire `IO#read` calls originating from the `Net::BER::BERParser` module to use non-blocking reads, depending on `IO.select` to block until the socket is readable or times out, raising a timeout error (instead of blocking forever; though that is still the default).
- [x] non-blocking reads
- [x] configurable timeout
- [ ] extensive unit & integration tests

The internal and external APIs are likely to change (feedback welcome). Would welcome general thoughts or concerns for these kinds of changes. From the recent research I've been doing, this seems to be the most reliable, forward-compatible way to implement non-blocking reads (using `IO.select`, as opposed to `setsockopt` with `SO_RCVTIMEO`).

Pairing this with similar changes for `write` would be the logical next step, assuming this makes sense. It's worth considering making non-blocking/blocking  calls configurable; this might be necessary since it's not always needed and is likely to behave in unexpected ways, so we'd not want that to be a surprise for anybody, or worse: have no recourse if it breaks their integration surprisingly.

I'm still thinking through what the end-result this would afford us, especially in terms of how the `Net::LDAP` API can change (`search_async`, for instance). This PR is pretty superficial (no major changes), but is a good step towards introducing more asynchronous, non-blocking API designs.

Thoughts?

see https://github.com/ruby-ldap/ruby-net-ldap/pull/164
cc @jch @schaary

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -21,6 +21,10 @@ class Net::LDAP::Connection #:nodoc:
       raise Net::LDAP::LdapError, "Connection to #{server[:host]} timed out."
     end
 
+    if server[:timeout]
+      @conn.read_ber_timeout = server[:timeout]
+    end
+
     if server[:encryption]
       setup_encryption server[:encryption]
     end


HarlemSquirrel:
@mtodd still looking to get this in?

PR #177, opened 2014-12-17
support a connection callback for proxies

ccutrer:
also refactors connection establishment to only go through one method


colmexdev:
Sorry to comment over this old pull request, but I've been having troubles to connect to LDAP through proxy (Nginx + Puma) with ECONNREFUSED error. I don't know why it's taking localhost as default socket, but I want to authenticate to a remote LDAP server and I can't find how is that possible given that it seems impossible to change host and port for connection.

   Any advice will be truly appreciated.

PR #178, opened 2014-12-18
Support for rfc3062 Password Modify, closes #163

### LINKED ISSUE #163: Support for rfc3062 Password Modify
Would be great if password modify was supported, perhaps something like:

```ruby
ldap = Net::LDAP.new(host: 127.0.0.1, port: 636, encryption: :ssl)
ldap.modify_password(dn: 'cn=Joe,ou=People', oldPassword: 'foo', newPassword: 'bar', auth: {})
                                     
```

http://tools.ietf.org/html/rfc3062

https://github.com/openldap/openldap/blob/master/clients/tools/ldappasswd.c

https://github.com/openldap/openldap/blob/master/include/ldap.h#L377

mynameisrufus:
This implements the password modify extended request
http://tools.ietf.org/html/rfc3062

Change existing password:

```ruby
dn = 'uid=modify-password-user1,ou=People,dc=rubyldap,dc=com'

auth = {
  method: :simple,
  username: dn,
  password: 'passworD1'
}

ldap.password_modify(
  dn: dn,
  auth: auth,
  old_password: 'passworD1',
  new_password: 'passworD2'
)
```

Or get the LDAP server to generate a password for you:

```ruby
dn = 'uid=modify-password-user1,ou=People,dc=rubyldap,dc=com'

auth = {
  method: :simple,
  username: dn,
  password: 'passworD1'
}

ldap.password_modify(
  dn: dn,
  auth: auth,
  old_password: 'passworD1'
)

ldap.get_operation_result.extended_response[0][0] #=> 'VtcgGf/G'
```

### DIFF
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -323,7 +323,14 @@ class Net::LDAP
     :constructed => constructed,
   }
 
+  universal = {
+    constructed: {
+      107 => :array #ExtendedResponse (PasswdModifyResponseValue)
+    }
+  }
+
   AsnSyntax = Net::BER.compile_syntax(:application => application,
+                                      :universal => universal,
                                       :context_specific => context_specific)
 
   DefaultHost = "127.0.0.1"
@@ -332,7 +339,8 @@ class Net::LDAP
   DefaultTreebase = "dc=com"
   DefaultForceNoPage = false
 
-  StartTlsOid = "1.3.6.1.4.1.1466.20037"
+  StartTlsOid = '1.3.6.1.4.1.1466.20037'
+  PasswdModifyOid = '1.3.6.1.4.1.4203.1.11.1'
 
   # https://tools.ietf.org/html/rfc4511#section-4.1.9
   # https://tools.ietf.org/html/rfc4511#appendix-A
@@ -651,8 +659,11 @@ class Net::LDAP
   #++
   def get_operation_result
     result = @result
-    result = result.result if result.is_a?(Net::LDAP::PDU)
     os = OpenStruct.new
+    if result.is_a?(Net::LDAP::PDU)
+      os.extended_response = result.extended_response
+      result = result.result
+    end
     if result.is_a?(Hash)
       # We might get a hash of LDAP response codes instead of a simple
       # numeric code.
@@ -1036,6 +1047,44 @@ class Net::LDAP
     instrument "modify.net_ldap", args do |payload|
       @result = use_connection(args) do |conn|
         conn.modify(args)
+      end
+      @result.success?
+    end
+  end
+
+  # Password Modify
+  #
+  # Change existing password:
+  #
+  #  dn = 'uid=modify-password-user1,ou=People,dc=rubyldap,dc=com'
+  #  auth = {
+  #    method: :simple,
+  #    username: dn,
+  #    password: 'passworD1'
+  #  }
+  #  ldap.password_modify(dn: dn,
+  #                       auth: auth,
+  #                       old_password: 'passworD1',
+  #                       new_password: 'passworD2')
+  #
+  # Or get the LDAP server to generate a password for you:
+  #
+  #  dn = 'uid=modify-password-user1,ou=People,dc=rubyldap,dc=com'
+  #  auth = {
+  #    method: :simple,
+  #    username: dn,
+  #    password: 'passworD1'
+  #  }
+  #  ldap.password_modify(dn: dn,
+  #                       auth: auth,
+  #                       old_password: 'passworD1')
+  #
+  #  ldap.get_operation_result.extended_response[0][0] #=> 'VtcgGf/G'
+  #
+  def password_modify(args)
+    instrument "modify_password.net_ldap", args do |payload|
+      @result = use_connection(args) do |conn|
+        conn.password_modify(args)
       end
       @result.success?
     end
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -539,6 +539,51 @@ class Net::LDAP::Connection #:nodoc:
     pdu
   end
 
+  ##
+  # Password Modify
+  #
+  # http://tools.ietf.org/html/rfc3062
+  #
+  # passwdModifyOID OBJECT IDENTIFIER ::= 1.3.6.1.4.1.4203.1.11.1
+  #
+  # PasswdModifyRequestValue ::= SEQUENCE {
+  #   userIdentity    [0]  OCTET STRING OPTIONAL
+  #   oldPasswd       [1]  OCTET STRING OPTIONAL
+  #   newPasswd       [2]  OCTET STRING OPTIONAL }
+  #
+  # PasswdModifyResponseValue ::= SEQUENCE {
+  #   genPasswd       [0]     OCTET STRING OPTIONAL }
+  #
+  # Encoded request:
+  #
+  #   00\x02\x01\x02w+\x80\x171.3.6.1.4.1.4203.1.11.1\x81\x100\x0E\x81\x05old\x82\x05new
+  #
+  def password_modify(args)
+    dn = args[:dn]
+    raise ArgumentError, 'DN is required' if !dn || dn.empty?
+
+    ext_seq = [Net::LDAP::PasswdModifyOid.to_ber_contextspecific(0)]
+
+    unless args[:old_password].nil?
+      pwd_seq = [args[:old_password].to_ber(0x81)]
+      pwd_seq << args[:new_password].to_ber(0x82) unless args[:new_password].nil?
+      ext_seq << pwd_seq.to_ber_sequence.to_ber(0x81)
+    end
+
+    request = ext_seq.to_ber_appsequence(Net::LDAP::PDU::ExtendedRequest)
+
+    message_id = next_msgid
+
+    write(request, nil, message_id)
+    pdu = queued_read(message_id)
+
+    if !pdu || pdu.app_tag != Net::LDAP::PDU::ExtendedResponse
+      raise Net::LDAP::ResponseMissingError, "response missing or invalid"
+    end
+
+    pdu
+  end
+
   #--
   # TODO: need to support a time limit, in case the server fails to respond.
   # Unlike other operation-methods in this class, we return a result hash
--- a/lib/net/ldap/pdu.rb
+++ b/lib/net/ldap/pdu.rb
@@ -74,6 +74,7 @@ class Net::LDAP::PDU
   attr_reader :search_referrals
   attr_reader :search_parameters
   attr_reader :bind_parameters
+  attr_reader :extended_response
 
   ##
   # Returns RFC-2251 Controls if any.
@@ -120,7 +121,7 @@ class Net::LDAP::PDU
     when UnbindRequest
       parse_unbind_request(ber_object[1])
     when ExtendedResponse
-      parse_ldap_result(ber_object[1])
+      parse_extended_response(ber_object[1])
     else
       raise LdapPduError.new("unknown pdu-type: #{@app_tag}")
     end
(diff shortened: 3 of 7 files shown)


net-ldap.gemspec
```diff
@@ -31,4 +31,5 @@ the most recent LDAP RFCs (4510-4519, plutions of 4520-4532).}
 
   s.add_development_dependency("flexmock", "~> 1.3")
   s.add_development_dependency("rake", "~> 10.0")
+  s.add_development_dependency("pry")
```
mtodd: :-1: as much as I like `pry`, I don't think this is a necessary development dependency.

test/fixtures/openldap/slapd.conf.ldif
```diff
@@ -3,7 +3,7 @@ objectClass: olcGlobal
 cn: config
 olcPidFile: /var/run/slapd/slapd.pid
 olcArgsFile: /var/run/slapd/slapd.args
-olcLogLevel: none
+olcLogLevel: -1
```
mtodd: :ok_hand:

test/integration/test_password_modify.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def test_password_modify
+    assert @ldap.password_modify(dn: @dn,
+                                 auth: @auth,
+                                 old_password: 'passworD1',
+                                 new_password: 'passworD2')
+
+    refute @ldap.get_gen_passwd_result, 'Should not have generated a new password'
+  end
```
mtodd: I would expect that we also test the new password works and the old password fails as a sanity check.

test/test_helper.rb
```diff
@@ -56,7 +56,7 @@ def setup
     @service = MockInstrumentationService.new
     @ldap = Net::LDAP.new \
       host:           ENV.fetch('INTEGRATION_HOST', 'localhost'),
-      port:           389,
+      port:           ENV.fetch('INTEGRATION_PORT', 389),
```
mtodd: :ok_hand:

lib/net/ldap/connection.rb
```diff
@@ -649,9 +689,7 @@ def add(args)
     write(request, nil, message_id)
     pdu = queued_read(message_id)
 
-    if !pdu || pdu.app_tag != Net::LDAP::PDU::AddResponse
-      raise Net::LDAP::LdapError, "response missing or invalid"
-    end
```
mtodd: :+1: I like this abstraction!

lib/net/ber/ber_parser.rb
```diff
@@ -10,7 +10,8 @@ module Net::BER::BERParser
     5 => :null,
     6 => :oid,
     10 => :integer,
-    13 => :string # (relative OID)
+    13 => :string, # (relative OID)
+    139 => :array
```
mtodd: Any relevant docs for this `139` value? (Cursory search doesn't bring up anything.)

lib/net/ldap.rb
```diff
@@ -645,6 +646,13 @@ def get_operation_result
     os
   end
 
+  # Returns the +genPasswd+ from the +PasswdModifyResponseValue+ after
+  # +password_modify+ operation has been called without specifying a new
+  # password.
+  def get_gen_passwd_result
+    @result.extended_response.nil? ? nil : @result.extended_response[0][0]
+  end
```
mtodd: Not a fan of this API. Should we be incorporating extended response results into `get_operation_result`?

lib/net/ldap.rb
```diff
@@ -1072,6 +1080,59 @@ def modify(args)
...
+      if @open_connection
+        @result = @open_connection.password_modify(args)
+      else
+        @result = 0
+        begin
+          conn = Connection.new \
+            :host                    => @host,
+            :port                    => @port,
+            :encryption              => @encryption,
+            :instrumentation_service => @instrumentation_service
```
mtodd: We've since merged https://github.com/ruby-ldap/ruby-net-ldap/pull/180 which provides a `use_connection` method to do this connection handling for you.

lib/net/ldap/connection.rb
```diff
@@ -622,10 +622,50 @@ def modify(args)
...
+  #   userIdentity    [0]  OCTET STRING OPTIONAL
+  #   oldPasswd       [1]  OCTET STRING OPTIONAL
+  #   newPasswd       [2]  OCTET STRING OPTIONAL }
+  #
+  # PasswdModifyResponseValue ::= SEQUENCE {
+  #   genPasswd       [0]     OCTET STRING OPTIONAL }
+  #
+  # Encoded request:
+  #
+  #   00\x02\x01\x02w+\x80\x171.3.6.1.4.1.4203.1.11.1\x81\x100\x0E\x81\x05old\x82\x05new
```
mtodd: :+1: thanks for this documentation!

lib/net/ldap/connection.rb
```diff
@@ -622,10 +622,50 @@ def modify(args)
...
+  #
+  def password_modify(args)
+    dn = args.fetch(:dn, -> { raise ArgumentError, 'Unable to modify empty DN' })
```
mtodd: We don't need a Lambda here. Thinking something simple like:

``` ruby
dn = args[:dn] || raise ArgumentError.new("DN is required")

# or
dn = args[:dn]
raise ArgumentError, "DN is required" if !dn || dn.empty?
```

lib/net/ldap/pdu.rb
```diff
@@ -181,6 +182,29 @@ def parse_ldap_result(sequence)
...
+  # Parse an extended response
+  #
+  # http://www.ietf.org/rfc/rfc2251.txt
+  #
+  # Each Extended operation consists of an Extended request and an
+  # Extended response.
+  #
+  #      ExtendedRequest ::= [APPLICATION 23] SEQUENCE {
+  #           requestName      [0] LDAPOID,
+  #           requestValue     [1] OCTET STRING OPTIONAL }
```
mtodd: :+1: great docs.

mtodd:
Thanks for submitting this!

My only gripe would be the `get_gen_passwd_result` API (as mentioned above). My reasoning is that our API should not be hardcoded for each extended operation possible but expose that data generally so new extensions don't require adding new methods to `Net::LDAP` et al. Does that make sense?

Caveat lector: I'm not intimately familiar with extended operations and how other APIs implement this. From taking a quick look at the Perl Net::LDAP library, its API is different enough such that returning a message object that responds to [`gen_password`](http://search.cpan.org/dist/perl-ldap/lib/Net/LDAP/Extension/SetPassword.pm) doesn't produce the same code smell. I'm in favor of emulating this API but that would require drastic changes to our own.

lib/net/ber/ber_parser.rb
mynameisrufus: This is most confusing part for me, I was hoping you could shed some light on it. This is the only way I could get it too work, I will have another look at it.

lib/net/ldap.rb
mynameisrufus: Agree

mynameisrufus:
Thanks for the great feedback, agree on all points.

lib/net/ber/ber_parser.rb
mtodd: @mynameisrufus tricky. I'm equally confused. Let me give it some thought and dig in a bit. Thanks for your patience on all this!

lib/net/ldap/connection.rb
```diff
@@ -691,10 +727,19 @@ def delete(args)
...
+
+  ##
+  # Raise +Net::LDAP::LdapError+ if the response is missing or an unexpected
+  # response tag is returned.
+
+  def assert_response!(pdu, tag)
+    return if pdu && pdu.app_tag == tag
+    raise Net::LDAP::LdapError, 'response missing or invalid'
+  end
+  private :assert_response!
```
jch: :+1: to this abstraction.

Couple of nitpicks:

* What's your reasoning for adding the `!` in `assert_response!`?
* [Minitest assertions](http://ruby-doc.org/stdlib-2.0/libdoc/minitest/rdoc/MiniTest/Assertions.html#method-i-assert_equal) typically have the expected value as the first argument, and the subject under test as the 2nd. It'd be good to flip the order of your argument list to match unless there's a good reason otherwise.
* Rather than raise an exception, you can use [`flunk`](http://ruby-doc.org/stdlib-2.0/libdoc/minitest/rdoc/MiniTest/Assertions.html#method-i-flunk) with a descriptive message. Something like:

```ruby
unless pdu && pdu.app_tag == tag
  flunk "Expected #{tag.inspect}, but got #{(pdu && pdu.app_tag).inspect}"
end
```

mtodd: @jch this is a runtime assertion, not in test. We were already enforcing this assertion, but it was repeated everywhere, hence the abstraction.

jch: Aaah, sorry. My brain automatically assumed it was a test helper. I guess the bang `!` is because it raises an exception.

mynameisrufus:
Rebased with feedback changes made, I have left the extended response a little raw as I think migrating to a better approach would require drastic changes like you said. Supporting auto generated password is a bit of a cherry on top anyway, I actually don't need it but figured implementing the entire operation as per the spec would be a win.

```ruby
ldap.get_operation_result.extended_response[0][0] #=> 'VtcgGf/G'
```

If I get time I will look around at other libs an see what a good way of doing this would be maybe open a separate PR.

mynameisrufus:
Any chance of getting this merged?

mtodd:
@mynameisrufus I'm really only blocked on researching the `139` primitive value, and haven't had the time to do that. If you'd like to dig in and get to the bottom of it, I'd really appreciate it. Otherwise hopefully the end of the week I'll have some time to check it out.

mynameisrufus:
I have moved the the value from primitive into `APPLICATION` see http://tools.ietf.org/html/rfc4511#section-4.10 for example that has `[APPLICATION 14]`, this seems incorrect as http://tools.ietf.org/html/rfc3062#section-2 does not have `[APPLICATION 43]` it's just a `SEQUENCE`. `0b01000000, # 64` + 43 gives us 139 the number that we had in primitive, this is only for the `genPasswd` response value if returned.

I could be wrong but other libs I have been reading do not really concern themselves with these types, they work more along the line of (python3-ldap):

```ruby
o = ModifyPassword.new(connection, options)
o.result #=> '2ex7c3'
```

Here it just assumes it's getting back a sequence?

mynameisrufus:
I have captured a response from wireshark:

```ruby
decode = ->(str) { str.split(':').map(&:hex).map(&:chr).join }
data = '30:1a:02:01:02:78:15:0a:01:00:04:00:04:00:8b:0c:30:0a:80:08:66:47:6d:63:75:62:34:71'

str = decode.call(data)

str #=> "0\x1A\x02\x01\x02x\x15\n\x01\x00\x04\x00\x04\x00\x8B\f0\n\x80\bfGmcub4q"

bytes = []

io = StringIO.new(str)

while byte = io.getbyte
  bytes << byte
end

bytes #=> [48, 26, 2, 1, 2, 120, 21, 10, 1, 0, 4, 0, 4, 0, 139, 12, 48, 10, 128, 8, 102, 71, 109, 99, 117, 98, 52, 113]

139.chr #=> "\x8B"
```

mynameisrufus:
@mtodd looks like ASCII 139 (8B in hex) is `‹` partial line forward http://www.fileformat.info/info/unicode/char/008b/index.htm

https://github.com/osstech-jp/openldap/blob/master/libraries/liblunicode/UnicodeData.txt#L140

The response should look like this:

```
0x
‹0
€fGmcub4q
```

I'm having trouble actually printing that out properly in irb :disappointed:

lib/net/ldap.rb
```diff
@@ -309,7 +309,14 @@ class Net::LDAP
     :constructed => constructed,
   }
 
+  universal = {
```
mynameisrufus: @mtodd I have moved the value into universal, context specific constructed will not work because it starts at 160 and it's not a primitive.

To be honest I don't know why this stuff is written the way it is when all that is resulted is an array with 256 length with values at index with the types, why not just one simple flat dict.

mtodd: @mynameisrufus I agree 100% that the parsing behavior needs to be refactored.

I'm uncomfortable accepting this as is without a better understanding of the specific value we're adding here. This change doesn't align with my understanding of the `ExtendedResponse` mechanics. I don't want to accept something that makes it work but isn't clear why because it will likely have unintended/unaccounted for consequences.

mynameisrufus: What do you mean by 'specific value'?, to me the specific value is in the title of the PR, support for modify password http://tools.ietf.org/html/rfc3062.

What is your understanding of the mechanics of `ExtendedResponse`, if you could share with me your understanding I would be happy to make changes to this PR.

I have invested considerable time in understanding this codebase and have addressed every point raised by yourself and @jch. I have sniffed packets, groked Open LDAP source and the source of many other libs in order to comprehensively document and explain each change I have made, I am committed to delivering this feature and like I said, I am happy to address any concerns you have or explain the workings of LDAP or this library. I fully understand that you have inherited this codebase so I understand the fear you may have in accepting changes to this lib.

mynameisrufus:
@jch @satoryu rebased :+1:

jch:
@mynameisrufus thanks for following through on this PR. I think you did a great job addressing all of our concerns. There's more cleanup we could do to the general API, but that's out of scope for your changes, and I think this feature has merit on it's own. I just released 0.13.0, so I will wait a bit in case there are other PR's to merge before cutting a new release.

Thanks!

mynameisrufus:
@jch happy new year and thank you.

mynameisrufus:
@jch happy new year and thank you!

Issue #179, opened 2014-12-18
Enforced code formatting


mynameisrufus:
I have been using https://github.com/bbatsov/rubocop in my code bases and am really happy with it. What are the chances of adding it to this one?

jch:
Added in #181

PR #181, opened 2014-12-19
Add rubocop to project, delivers #179

mynameisrufus:
This adds https://github.com/bbatsov/rubocop to the project with an auto
generated `.rubocop_todo.yml`. The violations can be addressed over time without
breaking the build.


mynameisrufus:
@mtodd I have added `pkg` to the ignore property in `.rubocop.yml` so you should be good now. Also a new task has been added called `rake ci` that will only run RuboCop on mri. I could not find anything to run a specific task for a specific ruby in the travis docs so implemented in the `Rakefile`. Should be good to merge now :fire:

jch:
:beers:

PR #183, opened 2015-01-04
Specific errors subclassing Net::LDAP::Error

satoryu:
I and my colleagues are looking forward to the release of @DenisKnauf's work #61 but unfortunately there were some conflicts while he was being busy :runner: 

so to take over the remaining work, I've recreated new pull request.


lib/net/ldap/error.rb
```diff
@@ -0,0 +1,31 @@
+class Net::LDAP
+  class Error < StandardError; end
```
jch: :+1: for switching the name from `LdapError` to just `Error`. Does anyone have strong opinions on having a backwards compatibility placeholder?

```ruby
class LdapError < Error
  def message
    "Deprecation warning: Net::LDAP::LdapError is no longer used. Use Net::LDAP::Error or rescue one of it's subclasses. [link to this file]\n" + super
  end
end
```

We're pre 1.0, there's no obligation for us to do this. I'm fine with either because it's an easy error to fix if anyone runs into it and we can document it in History.md

lib/net/ldap.rb
```diff
@@ -26,6 +26,7 @@ class LDAP
 require 'net/ldap/instrumentation'
 require 'net/ldap/connection'
 require 'net/ldap/version'
+require 'net/ldap/error'
```
jch: :+1: to putting this in it's own file

lib/net/ldap/connection.rb
```diff
@@ -12,13 +12,13 @@ def initialize(server)
...
+      raise Net::LDAP::Error, "No such address or other socket error."
     rescue Errno::ECONNREFUSED
-      raise Net::LDAP::LdapError, "Server #{server[:host]} refused connection on port #{server[:port]}."
+      raise Net::LDAP::Error, "Server #{server[:host]} refused connection on port #{server[:port]}."
     rescue Errno::EHOSTUNREACH => error
-      raise Net::LDAP::LdapError, "Host #{server[:host]} was unreachable (#{error.message})"
+      raise Net::LDAP::Error, "Host #{server[:host]} was unreachable (#{error.message})"
     rescue Errno::ETIMEDOUT
-      raise Net::LDAP::LdapError, "Connection to #{server[:host]} timed out."
+      raise Net::LDAP::Error, "Connection to #{server[:host]} timed out."
```
jch: Since we're making changes to this, I propose removing this rescue block entirely. The stdlib's `SocketError` is already descriptive about the specifics of the exception; We're not adding any additional information by wrapping it in our own custom error class. The change we would make is to document where the API may raise a `SocketError`.

lib/net/ldap/connection.rb
```diff
@@ -42,7 +42,7 @@ def close
   end
 
   def self.wrap_with_ssl(io, tls_options = {})
-    raise Net::LDAP::LdapError, "OpenSSL is unavailable" unless Net::LDAP::HasOpenSSL
+    raise Net::LDAP::Error, "OpenSSL is unavailable" unless Net::LDAP::HasOpenSSL
```
jch: Use `NoOpenSSLError `.

jch:
@satoryu nice start on this PR. Could you go over the changes and make sure that you're using the specific error when one is appropriate instead of using the generic `Net::LDAP::Error` class? Let me know when you have that done, and I can review again.

lib/net/ldap/password.rb
```diff
@@ -23,14 +23,14 @@ class << self
...
       when :md5
-        attribute_value = '{MD5}' + Base64.encode64(Digest::MD5.digest(str)).chomp!
+         attribute_value = '{MD5}' + Base64.encode64(Digest::MD5.digest(str)).chomp! 
       when :sha
-        attribute_value = '{SHA}' + Base64.encode64(Digest::SHA1.digest(str)).chomp!
+         attribute_value = '{SHA}' + Base64.encode64(Digest::SHA1.digest(str)).chomp! 
       when :ssha
-        salt = SecureRandom.random_bytes(16)
-        attribute_value = '{SSHA}' + Base64.encode64(Digest::SHA1.digest(str + salt) + salt).chomp!
+         srand; salt = (rand * 1000).to_i.to_s
```
mtodd: :-1: we need to continue using `SecureRandom` for this.

lib/net/ldap/connection.rb
mtodd: @jch definitely open to that, but let's tackle that separately.

lib/net/ldap/error.rb
mtodd: @jch I support the rename and keeping around the backwards compatible constant, though the semantics of the backwards compatible constant is a bit trickier than inheriting from `Error`. What needs to be backwards compatible is this:

``` ruby
begin
  # something that could fail
rescue Net::LDAP::LdapError => error
  # handle error
end
```

The problem with the above is that it wouldn't actually catch anything that inherits from `Net::LDAP::Error`, only the specific `Net::LDAP::LdapError` subclass. Instead, I think we can just have a constant assignment that would work as expected:

``` ruby
class Net::LDAP
  class Error < StandardError; end
  LdapError = Error
end

begin
rescue Net::LDAP::LdapError
  # should work
end
```

mtodd:
> make sure that you're using the specific error

I think the goal should be that we don't raise `Net::LDAP::Error` directly, always using the specific error class. Does that sound right, @jch?

lib/net/ldap/connection.rb
satoryu: @jch 's idea sounds good :+1: but I agree with @mtodd, the topic looks beyond the scope of this PR.

lib/net/ldap/password.rb
jch: Agree with @mtodd. This is also out of scope for this PR.

jch:
> I think the goal should be that we don't raise Net::LDAP::Error directly, always using the specific error class.

:+1: The base error class serves as a convenient catch-all for users to rescue.

lib/net/ldap/error.rb
jch: @mtodd good eye. What about having `Error` inherit from `LdapError` for a few point releases so that we can add a deprecation warning?

```ruby
class Net::LDAP
  class LdapError < StandardError
    def message
      "Deprecation warning... " + super
    end
  end
  
  class Error < LdapError; end
end
```

This allows us to eventually remove `LdapError`.

lib/net/ldap/connection.rb
jch: :metal:

lib/net/ldap/error.rb
mtodd: That would work!

lib/net/ldap/password.rb
mtodd: It may indicate that your branch was behind upstream `master`.

satoryu: :feelsgood: I miss it in merging.

lib/net/ldap/error.rb
satoryu: :+1:

satoryu:
@jch @mtodd I got the goal. I think there are 

- [ ] use `SecureRandom` as `salt` in `lib/net/ldap/password.rb`
- [ ] Define `Net::LDAP::LdapError` again and it says WARNING message. 
- [ ] Replace `Net::LDAP::Error` with appropriate error object

lib/net/ldap/password.rb
```diff
@@ -23,14 +23,14 @@ class << self
...
       when :md5
-        attribute_value = '{MD5}' + Base64.encode64(Digest::MD5.digest(str)).chomp!
+         attribute_value = '{MD5}' + Base64.encode64(Digest::MD5.digest(str)).chomp! 
       when :sha
-        attribute_value = '{SHA}' + Base64.encode64(Digest::SHA1.digest(str)).chomp!
+         attribute_value = '{SHA}' + Base64.encode64(Digest::SHA1.digest(str)).chomp! 
       when :ssha
-        salt = SecureRandom.random_bytes(16)
-        attribute_value = '{SSHA}' + Base64.encode64(Digest::SHA1.digest(str + salt) + salt).chomp!
+         srand; salt = SecureRandom.random_bytes(16)
```
jch: @satoryu Remove this dangling `stand`.

lib/net/ldap/connection.rb
```diff
@@ -412,10 +412,10 @@ def search(args = nil)
...
 
     # arg validation
-    raise Net::LDAP::LdapError, "search base is required" unless base
-    raise Net::LDAP::LdapError, "invalid search-size" unless size >= 0
-    raise Net::LDAP::LdapError, "invalid search scope" unless Net::LDAP::SearchScopes.include?(scope)
-    raise Net::LDAP::LdapError, "invalid alias dereferencing value" unless Net::LDAP::DerefAliasesArray.include?(deref)
+    raise Net::LDAP::NoSearchBaseError, "search base is required" unless base
+    raise Net::LDAP::SearchSizeInvalidError, "invalid search-size" unless size >= 0
+    raise Net::LDAP::SearchScopeInvalidError, "invalid search scope" unless Net::LDAP::SearchScopes.include?(scope)
+    raise Net::LDAP::Error, "invalid alias dereferencing value" unless Net::LDAP::DerefAliasesArray.include?(deref)
```
mtodd: We should either create a new exception for this or just use `ArgumentError`.

lib/net/ldap/connection.rb
mtodd: In the meantime, should we be using error classes for these errors? Or are we OK with keeping `Net::LDAP::Error` here?

lib/net/ldap/connection.rb
jch: :+1: to using `ArgumentError` for all of the above.

satoryu:
ops, Rubocop says that [Net::LDAP::Connection#search offenses](https://travis-ci.org/ruby-ldap/ruby-net-ldap/jobs/46504333#L461-L466)...
I'll try to fix it.

mtodd:
@satoryu looks like @mynameisrufus is working on a fix here: https://github.com/ruby-ldap/ruby-net-ldap/pull/185

satoryu:
:+1: good news.

jch:
I merged this manually after disabling rubocop in CI. We can continue to iterate on #185, but I wanted these changes in the release I'm cutting today (0.10.2).

satoryu:
@jch thanks!

PR #185, opened 2015-01-08
Extract paged search result cookie handling logic

mynameisrufus:
Extracted the rfc cookie find code in search into its own method to fix the
following offense that broke the build:

```
lib/net/ldap/connection.rb:379:3: C: Assignment Branch Condition size for search
is too high.
```

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -535,34 +535,9 @@ class Net::LDAP::Connection #:nodoc:
         payload[:page_count] ||= 0
         payload[:page_count]  += 1
 
-        # When we get here, we have seen a type-5 response. If there is no
-        # error AND there is an RFC-2696 cookie, then query again for the next
-        # page of results. If not, we're done. Don't screw this up or we'll
-        # break every search we do.
-        #
-        # Noticed 02Sep06, look at the read_ber call in this loop, shouldn't
-        # that have a parameter of AsnSyntax? Does this just accidentally
-        # work? According to RFC-2696, the value expected in this position is
-        # of type OCTET STRING, covered in the default syntax supported by
-        # read_ber, so I guess we're ok.
-        more_pages = false
-        if result_pdu.result_code == Net::LDAP::ResultCodeSuccess and controls
-          controls.each do |c|
-            if c.oid == Net::LDAP::LDAPControls::PAGED_RESULTS
-              # just in case some bogus server sends us more than 1 of these.
-              more_pages = false
-              if c.value and c.value.length > 0
-                cookie = c.value.read_ber[1]
-                if cookie and cookie.length > 0
-                  rfc2696_cookie[1] = cookie
-                  more_pages = true
-                end
-              end
-            end
-          end
-        end
-
-        break unless more_pages
+        cookie = next_search_result_cookie(result_pdu.result_code, controls)
+        break if cookie.nil?
+        rfc2696_cookie[1] = cookie
       end # loop
 
       # track total result count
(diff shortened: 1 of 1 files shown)


jch:
@mynameisrufus I thought rubocop was not going to cause build failures for existing code?

I think this could be a worthwhile refactoring, but I'm wary of doing a simple method extraction like this. Does this new method make sense standalone? Perhaps it does, but I think the method comment needs to be updated with a better explanation. Otherwise, we're just splitting and hiding behavior in more places and making the code harder to reason about.

mynameisrufus:
Sorry for late reply, I am on holiday.

I will try and figure out what this bit of code is for and change documentation.

mynameisrufus:
"we're just splitting and hiding behavior in more places and making the code harder to reason about" or is this encapsulating behaviour and making it easier to reason about?

jch:
@mynameisrufus I'm not saying your refactoring is hiding behavior. From your method name, it sounds like a good candidate for extraction, but the method comment should also work stand alone. The current description does not explain what a search cookie is, assumes there is a loop going on, and is generally tied to the caller. If I came across this, I would be confused:

> +  # When we get here, we have seen a type-5 response. If there is no
+  # error AND there is an RFC-2696 cookie, then query again for the next
+  # page of results. If not, we're done. Don't screw this up or we'll
+  # break every search we do.
+  #
+  # Noticed 02Sep06, look at the read_ber call in this loop, shouldn't
+  # that have a parameter of AsnSyntax? Does this just accidentally
+  # work? According to RFC-2696, the value expected in this position is
+  # of type OCTET STRING, covered in the default syntax supported by
+  # read_ber, so I guess we're ok.

mynameisrufus:
I have removed Austin Ziegler's 2010 code comments (8e9a8a48) and replaced it with a slimed down version with a link to the RFC, the RFC actually explains things, no point in repeating in comment.

jch:
:+1: nice. @mtodd care to chime in?

lib/net/ldap/connection.rb
```diff
@@ -535,34 +535,9 @@ def search(args = nil)
...
-                end
-              end
-            end
-          end
-        end
-
-        break unless more_pages
+        cookie = next_search_result_cookie(result_pdu.result_code, controls)
+        rfc2696_cookie[1] = cookie unless cookie.nil?
+        break if cookie.nil?
```
mtodd: Could reorder this and skip the `unless cookie.nil?`:

``` ruby
cookie = next_search_result_cookie(result.pdu.result_code, controls)
break if cookie.nil?
rfc2696_cookie[1] = cookie
```

lib/net/ldap/connection.rb
```diff
@@ -583,6 +558,33 @@ def search(args = nil)
...
+  #
+  # If there is no error AND there is an RFC-2696 cookie, then query again for
+  # the next page of results.
+
+  def next_search_result_cookie(result_code, controls)
+    found_cookie = nil
+    if result_code == Net::LDAP::ResultCodeSuccess and controls
+      controls.each do |c|
+        if c.oid == Net::LDAP::LDAPControls::PAGED_RESULTS
+          # just in case some bogus server sends us more than 1 of these.
```
mtodd: Does this comment apply anymore?

mtodd:
Only minor nits, this extraction looks great! Thanks @mynameisrufus!

mynameisrufus:
@mtodd made the changes, ready for merge :goat:

mtodd:
@mynameisrufus yep, looks great! I have some time tomorrow to do some manual testing (because I don't think our existing test suite covers paged search results) before merging.

PR #201, opened 2015-05-19
Add SHA2, :smd5, :sha1 & :ssha1 password generation support

ojab:
Haven't tested this PR on rubies < 2.2 yet, please consider it RFC.

With this patch net/ldap will support md5, sha, sha1, sha256, sha384, sha512 password schemes (and salted variants).

Given that comment says "I vote no because then  should you also provide ssha1 for symmetry?" -- should I remove :sha1/:ssha1 support?

Is the code overall looks ok for you or something should be changed?


mattcg:
@jch @mtodd is there anything holding this back? It'd be useful to have it merged!

mattcg:
Note that in this pull request, if the password contains accented characters, the following error will be thrown:

```
Encoding::CompatibilityError (incompatible character encodings: UTF-8 and ASCII-8BIT)
```

There's the option of forcing the encoding of the password to ASCII, but I think the more elegant solution is to use `digest.update`:

```ruby
    algo = Digest.module_eval(digest.upcase).new
    algo.update(str)
    algo.update(salt)
   "{#{type.upcase}}#{Base64.encode64(algo.digest + salt).chomp}"
```

jch:
@ojab @mattcg I think the idea is good, but I would prefer the changes to be more explicit and to have less metaprogramming. Comments inline.

lib/net/ldap/password.rb
```diff
@@ -6,33 +6,29 @@
...
-    attribute_value = ""
     def generate(type, str)
-      case type
-      when :md5
-         attribute_value = '{MD5}' + Base64.encode64(Digest::MD5.digest(str)).chomp!
-      when :sha
-         attribute_value = '{SHA}' + Base64.encode64(Digest::SHA1.digest(str)).chomp!
-      when :ssha
-         salt = SecureRandom.random_bytes(16)
-         attribute_value = '{SSHA}' + Base64.encode64(Digest::SHA1.digest(str + salt) + salt).chomp!
```
jch: Can we keep the same pattern here without introducing a KNOWN constant? Yes, there is more duplication, but it's simpler to reason about in my opinion and more obvious to grep for than hiding the information via lookups and `module_evals`.

HarlemSquirrel:
@ojab would you like to update this branch and get these other hash algorithms in here now?

PR #202, opened 2015-05-20
Fix Travis CI build

ojab:
gnutls-bin_3.0.11+really2.12.14-5ubuntu3.8_amd64.deb is gone from ubuntu repos, but travis attempts to download it. `apt-get update` should fix the issue.


mtodd:
Awesome, thanks @ojab!

PR #203, opened 2015-05-21
add Ruby 2.2.2 to .travis.yml

marutosi:

PR #205, opened 2015-05-22
Travis update

ojab:
…and another one PR about travis.
- Use ruby 2.1 instead of exact version 2.1.2
- Drop rbx-19mode
- Add ruby-2.2, ruby-head, jruby-head (last two are optional)

Feel free to reject this PR if this changes will be incorporated into other one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,12 @@ language: ruby
 rvm:
   - 1.9.3
   - 2.0.0
-  - 2.1.2
+  - 2.1
+  - 2.2
   # optional
+  - ruby-head
   - jruby-19mode
+  - jruby-head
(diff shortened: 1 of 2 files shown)


satoryu:
This conflicts with #203 . I think that this PR should be merged instead of #203 since this will provide the latest version of some series of ruby. how about @mtodd ?

mtodd:
@satoryu agreed.

@ojab this PR looks good! Could you rebase it off master? Conflict is minor and duplicated in this diff which should go away after a rebase.

ojab:
Done.

satoryu:
:+1:

mtodd:
Awesome, thank you!

PR #207, opened 2015-06-25
Fixing incorrect error class name

georgemillo:
This line in `connection.rb` tries to raise `Net::LDAP::ResponseMissingError` but that class doesn't exist (so I instead get a `NameError`).

I can only assume that it should be **`ResponseMissingOrInvalidError`**, as that's used elsewhere in this class, so I changed it.


mtodd:
@georgemillo good catch, thanks for submitting this :) I'll get this merged and released as soon as I can but I'm currently traveling so that might be later next week.

mtodd:
Still traveling so the release will come later in the week. Thanks again!

PR #211, opened 2015-08-25
Server strings are always UTF-8

cmdrclueless:
String values returned from an LDAP server are already encoded using UTF-8. Simple update the Ruby String so that it knows it's a UTF-8 value.

### DIFF
--- a/lib/net/ber.rb
+++ b/lib/net/ber.rb
@@ -296,9 +296,8 @@ end
 class Net::BER::BerIdentifiedString < String
   attr_accessor :ber_identifier
   def initialize args
+    args.force_encoding('UTF-8') if args.respond_to(:force_encoding)
     super args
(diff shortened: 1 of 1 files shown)


lib/net/ber.rb
```diff
@@ -296,9 +296,8 @@ def to_arr
 class Net::BER::BerIdentifiedString < String
   attr_accessor :ber_identifier
   def initialize args
+    args.force_encoding('UTF-8') if args.respond_to(:force_encoding)
     super args
-    # LDAP uses UTF-8 encoded strings
-    self.encode('UTF-8') if self.respond_to?(:encoding) rescue self
```
cmdrclueless: This line is effectively a noop. `self.encode` returns a new string and doesn't modify itself. To modify itself the call should have been `self.encode!`.

PR #212, opened 2015-08-25
Correctly set BerIdentifiedString values to UTF-8 

cmdrclueless:
The existing code did not set the actual internal value to UTF-8 when the encoding is correct. This patch attempts to resolve that problem by passing a UTF-8 encoded string to the super class when possible.

### DIFF
--- a/lib/net/ber.rb
+++ b/lib/net/ber.rb
@@ -296,8 +296,11 @@ end
 class Net::BER::BerIdentifiedString < String
   attr_accessor :ber_identifier
   def initialize args
-    args.force_encoding('UTF-8') if args.respond_to(:force_encoding)
-    super args
+    super begin
+      args.respond_to?(:encode) ? args.encode('UTF-8') : args
+    rescue
+      args
+    end
   end
 end
 
--- a/lib/net/ber.rb
+++ b/lib/net/ber.rb
@@ -296,9 +296,8 @@ end
 class Net::BER::BerIdentifiedString < String
   attr_accessor :ber_identifier
   def initialize args
+    args.force_encoding('UTF-8') if args.respond_to(:force_encoding)
     super args
-    # LDAP uses UTF-8 encoded strings
-    self.encode('UTF-8') if self.respond_to?(:encoding) rescue self
   end
 end
 


jch:
The change makes sense. Could you add a test for this?

I think it might be more readable to split this into two statements. It threw me off a bit to see `super begin`.

```ruby
encoded_args = begin
  args.respond_to?(:encode) ? args.encode('UTF-8') : args
rescue
  args
end

super(encoded_args)
```

Thoughts?

cmdrclueless:
@jch I would think the existing test handle the expectations on strings returned from the LDAP server. The problem is the original line
```
self.encode('UTF-8') if self.respond_to?(:encoding) rescue self
```
didn't really do anything. I *think* the intent of the code was to modify the string internally, so the correct call would have been `self.encode!`.  Is there a particular test that you're looking for?

As to the the change you propose, it doesn't really matter to me. If you prefer what you've proposed I can change it. Either one is valid.

jch:
I'm not sure whether the intention is to modify the string in place. It's weird to me that this inherits from String rather than composing it, but looking at [String::new](http://ruby-doc.org/core-2.2.0/String.html#method-c-new), the doc does say it strings a new String object. I agree with you that the current code doesn't make sense though because that value isn't used or saved anywhere. Where are you running into an error with this?

cmdrclueless:
I ran into this error when dealing with Chinese names in Active Directory.

cmdrclueless:
@jch, is there something more you're looking for before you a decision is made on this request?

jch:
@cmdrclueless I'm sorry for the slow response. Work has been busy, and the other maintainer is out of town. I think it's a good idea, I just want to test it more thoroughly myself before merging.

cmdrclueless:
@jch, no worries. I just wanted to make sure I had not dropped the ball on my side.

jch:
Did an audit through the codebase and the only callsite that initializes `BerIdentifiedString` is https://github.com/ruby-ldap/ruby-net-ldap/blob/4abe573ad0b920d94792cd939b78b93cad1dd91f/lib/net/ber/ber_parser.rb#L40. Seems like a reasonable fix. Thanks for the pull!

PR #213, opened 2015-08-27
Raise Net::LDAP::ConnectionRefusedError when new connection is refused.

satoryu:
Now Net::LDAP::Connection.new raises Net::LDAP::Error even if the
connection refused.

It's hard for some application to reconnect it only when refused.

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -14,7 +14,7 @@ class Net::LDAP::Connection #:nodoc:
     rescue SocketError
       raise Net::LDAP::Error, "No such address or other socket error."
     rescue Errno::ECONNREFUSED
-      raise Net::LDAP::Error, "Server #{server[:host]} refused connection on port #{server[:port]}."
+      raise Net::LDAP::ConnectionRefusedError, "Server #{server[:host]} refused connection on port #{server[:port]}."
     rescue Errno::EHOSTUNREACH => error
       raise Net::LDAP::Error, "Host #{server[:host]} was unreachable (#{error.message})"
     rescue Errno::ETIMEDOUT
--- a/lib/net/ldap/error.rb
+++ b/lib/net/ldap/error.rb
@@ -11,17 +11,18 @@ class Net::LDAP
   class SocketError < Error; end
   class ConnectionRefusedError < Error;
     def initialize(*args)
-      warn warning_message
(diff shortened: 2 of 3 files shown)


lib/net/ldap/connection.rb
```diff
@@ -14,7 +14,7 @@ def initialize(server)
     rescue SocketError
       raise Net::LDAP::Error, "No such address or other socket error."
     rescue Errno::ECONNREFUSED
-      raise Net::LDAP::Error, "Server #{server[:host]} refused connection on port #{server[:port]}."
+      raise Net::LDAP::ConnectionRefusedError, "Server #{server[:host]} refused connection on port #{server[:port]}."
```
satoryu: I came to think no need to wrap `Errno::ECONNREFUSED` in this situation. 
@mtodd @jch  How do you think?

jch: Personally, I prefer not wrapping the error because `Errno::ECONNREFUSED` is more descriptive and appropriate. In fact, I'd rather we didn't rescue any wrap any of the errors in this block. However, this would be a breaking change in the API and require changes to the caller. We're not at 1.0, but this does seem like a nasty surprise. How do you feel about making a PR that shows deprecation warnings?

satoryu: > How do you feel about making a PR that shows deprecation warnings?

sounds great to me :+1:

lib/net/ldap/error.rb
```diff
@@ -9,7 +9,22 @@ class Error < StandardError; end
...
+    end
+
+    def message
+      warn_deprecation_message
+      super
+    end
+
+    private
+    def warn_deprecation_message
+      warn "Deprecation warning: Net::LDAP::ConnectionRefused will be deprecated. Use Errno::ECONNREFUSED instead."
```
satoryu: When an user use this exception, this will show this exception will be removed. how about this? @jch

jch: This looks good. Could you add a test that [captures stderr](http://stackoverflow.com/questions/4459330/how-do-i-temporarily-redirect-stderr-in-ruby) and asserts that this message is being printed?

satoryu: Thanks, @jch ! I did it.

jch:
:tada: Awesome, thanks so much!

PR #217, opened 2015-09-17
obscure auth password upon #inspect, added test, closes #216

### LINKED ISSUE #216: `#inspect`-ing an instance of Net::LDAP leaks auth credentials (username and password instance variables)
Example:
```ruby
c = Net::LDAP.new(conn_hash)
p c
=> #<Net::LDAP:0x007ffc6698a9f8 
    @host="server-001", 
    @port=636, 
    @verbose=false, 
    @auth={:method=>:simple, :username=>"cn=user,o=org", :password=>"supersecret"}, 
    @base="dc=com", 
    @force_no_page=false, 
    @encryption={:method=>:simple_tls, :tls_options=>{}}, 
    @instrumentation_service=nil, 
    @open_connection=nil, 
    @result=#<Net::LDAP::PDU:0x007ffc669895f8 
    @message_id=1, 
    @app_tag=1, 
    @ldap_controls=[], 
    @ldap_result={:resultCode=>0, :matchedDN=>"", :errorMessage=>""}>>
```

This can be worked around when using this gem by wrapping Net::LDAP in another class, but I don't imagine this is desirable behavior.

Could something be done similar to GitHub's octokit gem ([highlighted LOC here](https://github.com/octokit/octokit.rb/blob/master/lib/octokit/client.rb#L109-L127)) to mask them (or remove entirely)?

astockwell:


jch:
Sweet! Thanks for including a test :+1:

lib/net/ldap.rb
```diff
@@ -1195,6 +1195,13 @@ def paged_searches_supported?
     @server_caps[:supportedcontrol].include?(Net::LDAP::LDAPControls::PAGED_RESULTS)
   end
 
+  # Mask auth password
+  def inspect
+    inspected = super
+    inspected.gsub! @auth[:password], "*******" if @auth[:password]
```
satoryu: :+1:

PR #219, opened 2015-09-18
Release v0.12

jch:
cc @ruby-ldap/ruby-ldap-admins

### DIFF
--- a/lib/net/ldap/version.rb
+++ b/lib/net/ldap/version.rb
@@ -1,5 +1,5 @@
 module Net
   class LDAP
-    VERSION = "0.11"
+    VERSION = "0.12"
   end
 end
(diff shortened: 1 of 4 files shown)


lib/net/ldap/version.rb
```diff
@@ -1,5 +1,5 @@
 module Net
   class LDAP
-    VERSION = "0.11"
+    VERSION = "0.12"
```
satoryu: Why isn't [Semantic Versioning](http://guides.rubygems.org/patterns/#semantic-versioning) adopted?

jch: @satoryu do you just mean why we're missing the patch number?

satoryu: Yes.

jch: Fixed in 98d122d3f94707e2367da8353ee2610ec095ff7c. Thanks!

satoryu: [LGTM](http://lgtm.in/i/SxXXctwTp)

jch:
@satoryu thanks for the review!

Issue #191, opened 2015-02-27
Support SRV record lookup and failover (rfc2782)


cperrin88:
To allow higher availability for LDAP, it would be nice if this implementation supported looking up SRV records according to rfc2782.

The lookup would be on 

    _ldap._tcp_.$HOSTNAME

and would return a list like

    _ldap._tcp.example.com. 600	IN	SRV	0 100 389 ldap1.example.com
    _ldap._tcp.example.com. 600	IN	SRV	0 100 389 ldap2.example.com
    _ldap._tcp.example.com. 600	IN	SRV	0 100 389 ldap3.example.com

These should be tried according to the priority and weight fields. Pages 5 and 6 of rfc2782 explain how te process should work.

Real world use would be for example for Active Directories in a HA setup.

mtodd:
@cperrin88 is this something that necessarily belongs inside of the client library or is something that can be negotiated externally and then passed into the client library once a host is selected?

cperrin88:
Yes I think so. It is a standard feature of Active Directory and this is the easiest way to make this centraly available. It is not uncommon for one DC to be down and Switching could be seamless without everyone putting that part in their software.

javanthropus:
I have created a gem that can assist with this feature request: https://rubygems.org/gems/resolv-srv.  The example included with its README file shows one way to use it with net-ldap to perform SRV record lookups, but it would be nicer if this facility were directly integrated into net-ldap.  That way the library could more seamlessly switch servers when necessary without the consumer code needing to handle the details.

jch:
@javanthropus thanks for the link. As valuable as this sounds, I don't have the bandwidth to maintain this. It looks like a straightforward single lib file. As a compromise, I propose adding a section to the readme that links to useful ruby LDAP utilities https://github.com/ruby-ldap/ruby-net-ldap/pull/220

javanthropus:
Thanks for adding the reference in the README.  Hopefully, others find the library and can make use of it.

I'm not sure what exactly you don't have bandwidth to maintain though.  If someone were to provide a PR with minimal changes that enables using this library as an external dependency, would you consider it?

It should be possible to add a single, optional attribute to Net::LDAP providing the domain in which to search that gets passed to Net::LDAP::Connection.  If that attribute is set Net::LDAP::Connection would then perform SRV lookups to find servers and try each one in turn until it successfully opens a connection.

javanthropus:
[...]

PR #223, opened 2015-09-28
Add the ability to provide a list of hosts for a connection

javanthropus:
This change adds the ability to provide a list of hosts to try when opening a connection while maintaining backward compatibility with existing code.  If a list is provided, it overrides the individual host and port settings; otherwise, the individual settings are used as a single element list.

The host list is a simple object that responds to `each` and provides a hostname and port pair to a block.  This allows for users to use any mechanism they need to provide a list of potential hosts.  For instance, a randomized ordering for a static list could be used, or a list based on SRV record lookups could be provided.

In either case, failure of all servers is required in order to fail opening a connection.  A `Net::LDAP::Error` instance that summarizes all of the server failures is raised in this case.  If any server succeeds, usage proceeds as usual.  This includes setting up encryption if requested.

### DIFF
--- a/test/test_ldap_connection.rb
+++ b/test/test_ldap_connection.rb
@@ -9,6 +9,44 @@ class TestLDAPConnection < Test::Unit::TestCase
     $stderr = stderr
   end
 
+  def test_list_of_hosts_with_first_host_successful
+    hosts = [
+              ['test.mocked.com', 636],
+              ['test2.mocked.com', 636],
+              ['test3.mocked.com', 636],
+            ]
+    flexmock(TCPSocket).should_receive(:new).ordered.with(*hosts[0]).once.and_return(nil)
+    flexmock(TCPSocket).should_receive(:new).ordered.never
+    Net::LDAP::Connection.new(:hosts => hosts.to_enum(:each))
+  end
+
+  def test_list_of_hosts_with_first_host_failure
+    hosts = [
+              ['test.mocked.com', 636],
+              ['test2.mocked.com', 636],
+              ['test3.mocked.com', 636],
+            ]
+    flexmock(TCPSocket).should_receive(:new).ordered.with(*hosts[0]).once.and_raise(SocketError)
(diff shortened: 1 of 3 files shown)


javanthropus:
I forgot to mention that this PR provides a way to tackle #191 that avoids pulling in any additional dependencies into this project.

I also have another change ready to go that DRYs up some of the connection creation and error handling logic between when sockets are given as opposed to host-port pairs.  I didn't include it here because it changes the messages of the error messages a bit, basically just wraps the original exception messages when raising `Net::LDAP::Error` and `Net::LDAP::ConnectionRefusedError` for them rather than include custom messages.

I can either add that commit here or apply it as a separate PR that builds on the result of accepting this PR so that it can be considered independently.  Let me know.

test/test_ldap_connection.rb
```diff
@@ -9,6 +9,44 @@ def capture_stderr
...
 
+  def test_list_of_hosts_with_first_host_successful
+    hosts = [
+              ['test.mocked.com', 636],
+              ['test2.mocked.com', 636],
+              ['test3.mocked.com', 636],
+            ]
+    flexmock(TCPSocket).should_receive(:new).ordered.with(*hosts[0]).once.and_return(nil)
+    flexmock(TCPSocket).should_receive(:new).ordered.never
+    Net::LDAP::Connection.new(:hosts => hosts.to_enum(:each))
```
jch: Why `to_enum` here? `hosts` is already an iterable object?

jch:
Looking good. Thanks for getting this started!

test/test_ldap_connection.rb
javanthropus: That was a late night edit, I think. I was trying to make sure that the
object passed in *only* responded to the `each` method so that the test
enforces the contract and doesn't allow the code to start using other
methods.

For now, I'll probably just chuck the `to_enum` calls unless you would
prefer the enforcement. It shouldn't too hard to make a simple wrapper
class to mask out the other methods though.

jch:
:tada:

PR #224, opened 2015-09-29
DRY up connection handling logic

javanthropus:
This patch simplifies configuration of new connections and removes some duplicate code around setting up encryption.  However, it does change the exception messages somewhat.  All the same exception types are raised in all the same places, but mostly the messages of underlying exceptions that `Net::LDAP::Error` and `Net::LDAP::ConnectionRefusedError` effectively wrap are used directly rather than providing new, different messages.  These changes don't break any tests, but it's possible that some user of the library may depend on the specific messages currently produced.

The upshot of this change of exception messages will mostly be felt when users provide host lists.  If all given hosts fail, each failure message for each host is included in an exception summarizing all of the failures along with the original exception type and host parameters that triggered the failure message.  This will certainly help when debugging failures in the multiple host case.

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -8,56 +8,54 @@ class Net::LDAP::Connection #:nodoc:
 
   def initialize(server)
     @instrumentation_service = server[:instrumentation_service]
-    server[:hosts] = [[server[:host], server[:port]]] if server[:hosts].nil?
 
     if server[:socket]
       prepare_socket(server)
     else
+      server[:hosts] = [[server[:host], server[:port]]] if server[:hosts].nil?
       open_connection(server)
     end
 
     yield self if block_given?
   end
 
-  def prepare_socket(server)
-    @conn = server[:socket]
+  def prepare_socket(server, close = false)
+    socket = server[:socket]
+    encryption = server[:encryption]
 
-    if server[:encryption]
-      setup_encryption server[:encryption]
-    end
+    @conn = socket
+    setup_encryption encryption if encryption
+  rescue
+    # Ensure the connection is closed when requested in the event of an SSL
+    # setup failure.
+    @conn.close if close
+    @conn = nil
+    raise
   end
 
   def open_connection(server)
+    hosts = server[:hosts]
+    encryption = server[:encryption]
+
     errors = []
-    server[:hosts].each do |host, port|
+    hosts.each do |host, port|
       begin
-        return connect_to_host(host, port, server)
-      rescue Net::LDAP::Error
-        errors << $!
+        prepare_socket(server.merge(socket: TCPSocket.new(host, port)), true)
+        return
+      rescue Net::LDAP::Error, SocketError, SystemCallError,
+             OpenSSL::SSL::SSLError
+        errors << [$!, host, port]
       end
     end
 
-    raise errors.first if errors.size == 1
+    if errors.size == 1
+      error = errors.first.first
+      raise Net::LDAP::ConnectionRefusedError, error.message if error.kind_of? Errno::ECONNREFUSED
+      raise Net::LDAP::Error, error.message
+    end
+
     raise Net::LDAP::Error,
-      "Unable to connect to any given server: \n  #{errors.join("\n  ")}"
-  end
-
-  def connect_to_host(host, port, server)
-    begin
-      @conn = TCPSocket.new(host, port)
-    rescue SocketError
-      raise Net::LDAP::Error, "No such address or other socket error."
-    rescue Errno::ECONNREFUSED
-      raise Net::LDAP::ConnectionRefusedError, "Server #{host} refused connection on port #{port}."
-    rescue Errno::EHOSTUNREACH => error
-      raise Net::LDAP::Error, "Host #{host} was unreachable (#{error.message})"
-    rescue Errno::ETIMEDOUT
-      raise Net::LDAP::Error, "Connection to #{host} timed out."
-    end
-
-    if server[:encryption]
-      setup_encryption server[:encryption]
-    end
+      "Unable to connect to any given server: \n  #{errors.map { |e, h, p| "#{e.class}: #{e.message} (#{h}:#{p})" }.join("\n  ")}"
   end
 
   module GetbyteForSSLSocket
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -48,14 +48,7 @@ class Net::LDAP::Connection #:nodoc:
       end
     end
 
-    if errors.size == 1
-      error = errors.first.first
-      raise Net::LDAP::ConnectionRefusedError, error.message if error.kind_of? Errno::ECONNREFUSED
-      raise Net::LDAP::Error, error.message
-    end
-
-    raise Net::LDAP::Error,
-      "Unable to connect to any given server: \n  #{errors.map { |e, h, p| "#{e.class}: #{e.message} (#{h}:#{p})" }.join("\n  ")}"
+    raise Net::LDAP::ConnectionError.new(errors)
   end
 
   module GetbyteForSSLSocket
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -34,14 +34,12 @@ class Net::LDAP::Connection #:nodoc:
     errors = []
     hosts.each do |host, port|
       begin
-        socket = TCPSocket.new(host, port)
-        prepare_socket(server.merge(socket: socket))
+        prepare_socket(server.merge(socket: TCPSocket.new(host, port)))
         return
       rescue Net::LDAP::Error, SocketError, SystemCallError,
              OpenSSL::SSL::SSLError => e
         # Ensure the connection is closed in the event a setup failure.
-        socket.close unless socket.nil?
-        socket = nil
+        close
         errors << [e, host, port]
       end
     end
@@ -145,6 +143,7 @@ class Net::LDAP::Connection #:nodoc:
   # have to call it, but perhaps it will come in handy someday.
   #++
   def close
+    return if @conn.nil?
     @conn.close
     @conn = nil
   end
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -19,18 +19,12 @@ class Net::LDAP::Connection #:nodoc:
     yield self if block_given?
   end
 
-  def prepare_socket(server, close = false)
+  def prepare_socket(server)
     socket = server[:socket]
     encryption = server[:encryption]
 
     @conn = socket
     setup_encryption encryption if encryption
-  rescue
-    # Ensure the connection is closed when requested in the event of an SSL
-    # setup failure.
-    @conn.close if close
-    @conn = nil
-    raise
   end
 
   def open_connection(server)
@@ -40,10 +34,14 @@ class Net::LDAP::Connection #:nodoc:
     errors = []
     hosts.each do |host, port|
       begin
-        prepare_socket(server.merge(socket: TCPSocket.new(host, port)), true)
+        socket = TCPSocket.new(host, port)
+        prepare_socket(server.merge(socket: socket))
         return
       rescue Net::LDAP::Error, SocketError, SystemCallError,
              OpenSSL::SSL::SSLError => e
+        # Ensure the connection is closed in the event a setup failure.
+        socket.close unless socket.nil?
+        socket = nil
         errors << [e, host, port]
       end
     end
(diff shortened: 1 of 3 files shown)


lib/net/ldap/connection.rb
```diff
@@ -8,56 +8,49 @@ class Net::LDAP::Connection #:nodoc:
...
+  rescue
+    # Ensure the connection is closed when requested in the event of an SSL
+    # setup failure.
+    @conn.close if close
+    @conn = nil
+    raise
   end
 
-  def open_connection(server)
+  def open_connection(hosts, encryption)
```
jch: I know this is a private class, but why change the method signature for `open_connection`? All of the information is available in the previous `server` hash.

lib/net/ldap/connection.rb
```diff
@@ -8,56 +8,49 @@ class Net::LDAP::Connection #:nodoc:
...
     yield self if block_given?
   end
 
-  def prepare_socket(server)
-    @conn = server[:socket]
-
-    if server[:encryption]
-      setup_encryption server[:encryption]
-    end
+  def prepare_socket(socket, encryption, close = false)
```
jch: Same question as `open_connection`.

lib/net/ldap/connection.rb
```diff
@@ -8,56 +8,49 @@ class Net::LDAP::Connection #:nodoc:
...
-    rescue Errno::ECONNREFUSED
-      raise Net::LDAP::ConnectionRefusedError, "Server #{host} refused connection on port #{port}."
-    rescue Errno::EHOSTUNREACH => error
-      raise Net::LDAP::Error, "Host #{host} was unreachable (#{error.message})"
-    rescue Errno::ETIMEDOUT
-      raise Net::LDAP::Error, "Connection to #{host} timed out."
+    if errors.size == 1
+      error = errors.first.first
+      raise Net::LDAP::ConnectionRefusedError, error.message if error.kind_of? Errno::ECONNREFUSED
+      raise Net::LDAP::Error, error.message
```
jch: This smells a bit. What prevents us from processing these errors the same way you do above around line 38?

jch:
> The upshot of this change of exception messages will mostly be felt when users provide host lists. If all given hosts fail, each failure message for each host is included in an exception summarizing all of the failures along with the original exception type and host parameters that triggered the failure message

This sounds like a worthwhile goal, but your PR is also trying to clean up code at the same time. Could you minimize your changeset to focus on this grouping of errors?

lib/net/ldap/connection.rb
javanthropus: This is mostly to make it explicit what arguments these methods actually use.  Passing around a hash makes it necessary to inspect the guts of the method to know what is really needed.

lib/net/ldap/connection.rb
javanthropus: This definitely smells, but it's mostly to ensure that we still return the same error types when using only a single host rather than a list.  I also saw notes in the code stating that the `ConnectionRefusedError` exception is deprecated, so this seems like a reasonable way to keep it around for just a little while longer.  Once that deprecated exception is removed, this logic will be simplified a bit.

The other thing done in this section is avoiding the need to define our own error messages.  They don't really provide any useful information beyond what the original exceptions provided.  For the single host case, we don't really need to add any notes about the server that failed since that's implicit to the app.  For the multiple host case, the new summary error includes each server's hostname and port in a consistent way regardless of the specific error type.

lib/net/ldap/connection.rb
```diff
@@ -8,56 +8,49 @@ class Net::LDAP::Connection #:nodoc:
...
-    server[:hosts] = [[server[:host], server[:port]]] if server[:hosts].nil?
 
     if server[:socket]
-      prepare_socket(server)
+      prepare_socket(server[:socket], server[:encryption])
     else
-      open_connection(server)
+      hosts = server[:hosts]
+      hosts = [[server[:host], server[:port]]] if hosts.nil?
+      open_connection(hosts, server[:encryption])
```
jch: I should've suggested this in your previous PR, but what do you think of this idea: Instead of having `open_connection` worry about handling multiple hosts and then checking whether it did one host, or multiple, we can iterate through `server[:hosts]` here. The advantage is that `open_connection` would only be responsible for what it's method name suggests: opening a single connection. You can then collect the errors here and handle it in a generic way without the code smell we're discussing below. It'd also minimize remove the changes from `open_connection`, which I think make the method less readable.

lib/net/ldap/connection.rb
jch: :+1: that's a good reason, but could you revert that and open it in a follow up PR?

It makes it easier to revisit merged PR's linked from the changelog.

lib/net/ldap/connection.rb
```diff
@@ -8,56 +8,49 @@ class Net::LDAP::Connection #:nodoc:
...
+      error = errors.first.first
+      raise Net::LDAP::ConnectionRefusedError, error.message if error.kind_of? Errno::ECONNREFUSED
+      raise Net::LDAP::Error, error.message
     end
 
-    if server[:encryption]
-      setup_encryption server[:encryption]
-    end
+    raise Net::LDAP::Error,
+      "Unable to connect to any given server: \n  #{errors.map { |e, h, p| "#{e.class}: #{e.message} (#{h}:#{p})" }.join("\n  ")}"
```
jch: Could you add a new test for this?

lib/net/ldap/connection.rb
javanthropus: I can do that.  Which would you like to tackle first, the PR that simplifies parameter handling or this one?

lib/net/ldap/connection.rb
javanthropus: I think that's a good idea.  I'll look into it.

lib/net/ldap/connection.rb
javanthropus: The existing tests for handling multiple hosts should already cover this, unless you want a test to check the error message itself rather than just the exception class.  Checking the message would be brittle, so I would recommend against it.

lib/net/ldap/connection.rb
jch: Your call. Whatever's easier for you.

javanthropus:
Before I start working on things here, it sounds like you would like to break this into 3 PRs: 1 that changes parameter handling in the `open_connection` and `prepare_socket` methods, 1 that reworks exception handling, and then whatever is left goes into this one.  Does that sound right?

jch:
Just one PR for changing the method signature, and the rest in this PR is fine.

javanthropus:
Cool deal.  I'll see what I can do this evening.

lib/net/ldap/connection.rb
javanthropus: After looking into this, I think it won't work out.  Most of what `open_connection` does in this patch is manage exception dispatching, and it needs to do this because in the multi-host case it must suppress any exceptions raised if a connection is successfully opened.  Opening the connection is actually handled by `prepare_socket` and `TCPSocket.new`, one line in the `open_connection` method.  The rest of the method is the looping over hosts to try, gathering of exceptions, and finally raising an exception based on the captured exceptions if no connection was opened.

In other words, to do what you suggest moves about 19 lines from `open_connection` into `initialize`, leaving `open_connection` as a one-liner and `initialize` rather more bloated.  I don't really see how this would address the exception handling code smell either since we're just moving logic around rather than eliminating or refactoring it significantly.

Have you looked at the result of the patch rather than the diffs yet?  I think you'll see that the patch is noisy only because it's simplifying the code.

lib/net/ldap/connection.rb
javanthropus: I believe I better understand what you were asking about here.  We could only avoid lines 48 and 49 by creating and recording the LDAP wrapper exceptions preemptively at line 42 rather than directly recording the raw exceptions in the errors array.  However, this is wasteful in the multi-host case due to the creation of additional exception objects that would be discarded in the more likely event that a secondary host actually succeeds.

A better solution might be to create a new exception class that takes the errors array as an argument.  It could basically wrap up the logic from lines 46 through 53.  By overriding the `new` class method on that class, we can return the expected exception class for the single error case and only create the new class' instance in the multi-error case.  This would also neatly hide the message string creation nicely.  Should I try for that in this PR or keep things as they are for now and do that in a new PR?

javanthropus: @jch I created a new exception `Net::LDAP::ConnectionError` to encapsulate this smell.  The idea here would be to eventually remove the smelly bits (currently located in the overridden `new` method) and simply use this exception both for the single host case as well as the multi-host case.

I haven't committed this change yet because I want to know if switching to this exception for connection failures make sense in the future.  It's a descendant of `Net::LDAP::Error`, so consumer incompatibilities should be minimized.  What do you think?

javanthropus:
@jch I have rewritten the branch for this pull request to hopefully address the main concerns raised regarding exception handling and method signature changes.  We can't really escape the code smell of the exception handling here, but it should be easier to clean up in the future if we allow for slightly different exceptions to be raised for the single host connection attempt.

lib/net/ldap/connection.rb
```diff
@@ -8,56 +8,47 @@ class Net::LDAP::Connection #:nodoc:
...
 
-    if server[:encryption]
-      setup_encryption server[:encryption]
-    end
+    @conn = socket
+    setup_encryption encryption if encryption
+  rescue
+    # Ensure the connection is closed when requested in the event of an SSL
+    # setup failure.
+    @conn.close if close
```
jch: This seems like new behavior. Why this change in this PR?

lib/net/ldap/connection.rb
```diff
@@ -8,56 +8,47 @@ class Net::LDAP::Connection #:nodoc:
...
+    hosts.each do |host, port|
       begin
-        return connect_to_host(host, port, server)
-      rescue Net::LDAP::Error
-        errors << $!
+        prepare_socket(server.merge(socket: TCPSocket.new(host, port)), true)
+        return
+      rescue Net::LDAP::Error, SocketError, SystemCallError,
+             OpenSSL::SSL::SSLError
+        errors << [$!, host, port]
```
jch: Assign the error to a variable rather than using the implicit here:

```ruby
rescue Net::LDAP::Error, SocketError, SystemCallError, OpenSSL::SSL::SSLError => e
  errors << [e, host, port]
end
```

jch:
Thanks for picking this up again

lib/net/ldap/connection.rb
javanthropus: Will do.

lib/net/ldap/connection.rb
javanthropus: Sockets created by the library in `open_connection` need to be closed in the event that they cannot successfully be configured, such as for encryption.  Otherwise, we leak file descriptors and hope that the GC releases them in a timely manner.  Not good, especially in the multiple host case where we may fail to configure the sockets for numerous hosts in rapid succession.

This can also be handled within the `open_connection` method since that is the only place where we currently set the `close` parameter.  Now that you bring it up, I think I like this option better, so I'll make this change.

javanthropus:
@jch What do you think about these latest changes?

lib/net/ldap/connection.rb
```diff
@@ -8,11 +8,11 @@ class Net::LDAP::Connection #:nodoc:
 
   def initialize(server)
     @instrumentation_service = server[:instrumentation_service]
-    server[:hosts] = [[server[:host], server[:port]]] if server[:hosts].nil?
 
     if server[:socket]
       prepare_socket(server)
     else
+      server[:hosts] = [[server[:host], server[:port]]] if server[:hosts].nil?
```
jch: Note: This works even when `server[:host]` and/or `server[:port]` are `nil`. It raises a `SocketError` and matches up with the old behavior.

lib/net/ldap/connection.rb
```diff
@@ -20,44 +20,33 @@ def initialize(server)
...
-      rescue Net::LDAP::Error
-        errors << $!
+        socket = TCPSocket.new(host, port)
+        prepare_socket(server.merge(socket: socket))
+        return
+      rescue Net::LDAP::Error, SocketError, SystemCallError,
+             OpenSSL::SSL::SSLError => e
+        # Ensure the connection is closed in the event a setup failure.
+        socket.close unless socket.nil?
+        socket = nil
```
jch: Could you move the nil check to `Connection#close` and call that?

jch:
@javanthropus thanks for the ping. I added one last piece of small feedback. We're getting close!

javanthropus:
@jch Sorry for the delay on getting the last changes in.  Let me know how this looks.

jch:
@javanthropus great work, thanks!

PR #225, opened 2015-09-29
add slash to attribute value filter

rlanore:
A filter like departmentNumber=FOO/BAR raise invalid filter syntax.
This PR fix this.


jch:
@rlanore Thanks for the pull!

PR #226, opened 2015-10-03
Define auth adapters

satoryu:
This PR provide a scheme of defining authentication (bind) of LDAP. 
By merging this PR, we'll be able to define new auth adapter. 

1. create new file can be `require`d as `'net/ldap/auth_adapter/#{auth scheme name}`
2. define an adapter as a subclass of `Net::LDAP::AuthAdapter` 
3. register new adapter by `Net::LDAP::AuthAdapter.register`

I hope this PR would help #210 :)


lib/net/ldap/auth_adapters/gss_spnego.rb
```diff
@@ -0,0 +1,42 @@
...
+      #
+      # This authentication method is accessed by calling #bind with a :method
+      # parameter of :gss_spnego. It requires :username and :password
+      # attributes, just like the :simple authentication method. It performs a
+      # GSS-SPNEGO authentication with the server, which is presumed to be a
+      # Microsoft Active Directory.
+      #++
+      class GSS_SPNEGO < Net::LDAP::AuthAdapter
+        def bind(auth)
+          require 'ntlm'
```
satoryu: This adapter depends on `ntlm` as the above comments point out.  now this can be defined as a gem like `net-ldap-auth_adapters-gss_spnego`.

lib/net/ldap/auth_adapters/simple.rb
```diff
@@ -0,0 +1,36 @@
...
+
+          raise Net::LDAP::BindingInformationInvalidError, "Invalid binding information" unless (user && psw)
+
+          message_id = @connection.next_msgid
+          request    = [
+            Net::LDAP::Connection::LdapVersion.to_ber, user.to_ber,
+            psw.to_ber_contextspecific(0)
+          ].to_ber_appsequence(Net::LDAP::PDU::BindRequest)
+
+          @connection.send(:write, request, nil, message_id)
```
satoryu: Calling a private method looks dirty... do you have any idea?

lib/net/ldap/auth_adapters/anonymous.rb
```diff
@@ -0,0 +1,3 @@
+require 'net/ldap/auth_adapters/simple'
+
+Net::LDAP::AuthAdapter.register(:anonymous, Net::LDAP::AuthAdapters::Simple)
```
jch: Why have both `:anon` and `:anonymous`? Is this for backwards compatibility?

lib/net/ldap/auth_adapters/simple.rb
jch: I agree, but this is the existing implementation, so let's change this in a separate PR.

lib/net/ldap/connection.rb
```diff
@@ -250,131 +250,13 @@ def next_msgid
...
-      psw.to_ber_contextspecific(0)
-    ].to_ber_appsequence(Net::LDAP::PDU::BindRequest)
-
-    write(request, nil, message_id)
-    pdu = queued_read(message_id)
-
-    if !pdu || pdu.app_tag != Net::LDAP::PDU::BindResult
-      raise Net::LDAP::NoBindResultError, "no bind result"
+      require "net/ldap/auth_adapters/#{meth}"
+      adapter = Net::LDAP::AuthAdapter[meth]
```
jch: Should test and handle the case when an adaptor is not found.

lib/net/ldap/auth_adapters/anonymous.rb
jch: Ah, I see the previous case statement works with `:simple, :anonymous, :anon`

lib/net/ldap/connection.rb
```diff
@@ -250,131 +250,13 @@ def next_msgid
...
-      LdapVersion.to_ber, user.to_ber,
-      psw.to_ber_contextspecific(0)
-    ].to_ber_appsequence(Net::LDAP::PDU::BindRequest)
-
-    write(request, nil, message_id)
-    pdu = queued_read(message_id)
-
-    if !pdu || pdu.app_tag != Net::LDAP::PDU::BindResult
-      raise Net::LDAP::NoBindResultError, "no bind result"
+      require "net/ldap/auth_adapters/#{meth}"
```
jch: This works, but I prefer to have `requires` at the top of a file to make it obvious. Also, until we extract the other adaptors into gems, I'd suggest putting these requires in [`lib/net/ldap.rb`](https://github.com/ruby-ldap/ruby-net-ldap/blob/master/lib/net/ldap.rb#L20-L29):

```ruby
require 'net/ldap/auth_adapter'
require 'net/ldap/auth_adapter/simple'  # <-- side note, can we rename the folder and namespace to be singular?
require 'net/ldap/auth_adapter/sasl'
# ...

# Move all of the registration to a single place, decouples the definition of the adapter from it's registration and use
Net::LDAP::AuthAdapter.register([:simple, :anon, :anonymous], Net::LDAP::AuthAdapters::Simple)
Net::LDAP::AuthAdapter.register(:sasl, Net::LDAP::AuthAdapters::Sasl)
```

lib/net/ldap/auth_adapters/gss_spnego.rb
```diff
@@ -0,0 +1,42 @@
+require 'net/ldap/auth_adapter'
+require 'net/ldap/auth_adapters/sasl'
```
jch: Why do you need sasl here?

jch:
@satoryu thanks for getting this started! It's looking good.

lib/net/ldap/auth_adapters/simple.rb
satoryu: I got. thanks.

lib/net/ldap/connection.rb
```diff
@@ -250,131 +250,13 @@ def next_msgid
...
-    raise Net::LDAP::BindingInformationInvalidError, "Invalid binding information" unless (user && psw)
-
-    nego = proc { |challenge|
-      t2_msg = NTLM::Message.parse(challenge)
-      t3_msg = t2_msg.response({ :user => user, :password => psw },
-                               { :ntlmv2 => true })
-      t3_msg.serialize
-    }
-
-    bind_sasl(:method => :sasl, :mechanism => "GSS-SPNEGO",
```
satoryu: > Why do you need sasl here?

this is because the current version of `bind_gss_snego` uses `bind_sasl`.

satoryu:
@jch thank you for your comments :) I've fixed all you pointed out. so please check this PR again.

lib/net/ldap/connection.rb
jch: :+1:

test/test_auth_adapter.rb
```diff
@@ -0,0 +1,11 @@
+require 'test_helper'
+
+class TestAuthAdapter < Test::Unit::TestCase
+  def test_undefined_auth_adapter
+    flexmock(TCPSocket).should_receive(:new).ordered.with('ldap.example.com', 379).once.and_return(nil)
```
jch: It's too bad that `Connection.new` actually tries to open the socket and forces us to stub here. Not suggesting any change, but making an observation.

lib/net/ldap.rb
```diff
@@ -27,6 +27,14 @@ class LDAP
 require 'net/ldap/connection'
 require 'net/ldap/version'
 require 'net/ldap/error'
+require 'net/ldap/auth_adapter'
+require 'net/ldap/auth_adapter/simple'
+require 'net/ldap/auth_adapter/sasl'
+require 'net/ldap/auth_adapter/gss_spnego'
```
jch: Can we leave this adapter out of this PR since it's not one of the existing adapters we currently support? After this merges, we can create a separate gem and link to it from the README.

test/test_auth_adapter.rb
satoryu: Thank you for your feedback. I had no idea so I referred to `test/test_ldap_connection.rb`
What would you do if you were me in this case?

jch:
@satoryu thanks for following through with this PR. Would you be interested in making a separate gem for `gss_spnego`? An alternative to a separate gem is to just document it since there isn't that much code.

satoryu:
@jch Yes. but give me time to implement the separate gem. I guess I would bother much time since there are no test code for `bind_gss_spnego`... :feelsgood: 

BTW, Is the release of v0.12.0 blocked by providing this separate gem?

jch:
@satoryu not blocked. I don't believe there's any breaking changes.

satoryu:
:+1:

Issue #117, opened 2014-10-09
Do we want to support method style attribute readers for Net::LDAP::Entry?


jch:
The current implementation of `Net::LDAP::Entry#valid_argument?` requires an attribute to exist before you can use the method style accessors:

``` ruby
entry = Net::LDAP::Entry.new
entry.respond_to? :sn  # false
entry.sn  # raises NoMethodError

entry[:sn] = "foo"
entry.respond_to? :sn  # true
entry.sn = "bar"
entry.sn   # => "bar"
entry["sn"]  # => "bar"
entry["Sn"] # => "bar"
```

This feels inconsistent and confusing. For 1.0, I'd like to clearly define how attributes can be accessed.

I looked into [RFC 4517](http://tools.ietf.org/html/rfc4517) for rules about valid attribute names, but didn't find any.
- Are attributes case sensitive?
- Can they have any characters in them? 
- Do we want to support so many different styles of access?

cc @ruby-ldap/ruby-ldap-admins

yairgo:
I'm kind of confused by what is being suggested here. If I specify in my search to only return these values: [:sn, :mail, :samaccountname], are you suggesting that if i call entry.memberof that it should just return nil?

PR #227, opened 2015-10-16
Implement to_h method for entry

kremez19:
Hi guys!

I think it makes sense to have publicly available `to_h` method instead of dangerous `instance_variable_get` usage when we need actually hash, not an object, isn't it?


jch:
@kremez19 thanks for the pr! I'm not a fan of the current implementation of `[]` for accessing attributes, but I don't think introducing another public accessor is the way to fix it. https://github.com/ruby-ldap/ruby-net-ldap/issues/117 has some additional background if you're interested.

I'm going to close this PR because I don't want to add this interface, but please feel free to keep the discussion going.

PR #232, opened 2015-11-05
Set operation result if LDAP server is not accessible

### LINKED ISSUE #231: Bind Error is reported as success
I used version 0.12.0. No LDAP-server running! So the bind fails.

Given that script:

~~~ruby
#!/usr/bin/env ruby

require 'net/ldap'

ldap_client = Net::LDAP.new(host: '127.0.0.1', port: '12345', base: 'dc=example,dc=org')
begin
  ldap_client.bind(method: :simple, username: 'asdf', password: 'asdf')
rescue => e
  $stderr.puts e.class
  $stderr.puts e.message
  $stderr.puts ldap_client.get_operation_result
  $stderr.puts ldap_client.get_operation_result.message
end

~~~

When you save it as `my-client.rb` and run it, it outputs `success`. Not what I expected.

~~~bash
ruby my-script
# => Deprecation warning: Net::LDAP::ConnectionRefused will be deprecated. Use Errno::ECONNREFUSED instead.
# => Net::LDAP::ConnectionRefusedError
# => Deprecation warning: Net::LDAP::ConnectionRefused will be deprecated. Use Errno::ECONNREFUSED instead.
# => Connection refused - connect(2) for "127.0.0.1" port 12345
# => #<OpenStruct code=0, message="Success">
# => Success
~~~

I would have expected it to return `ResultCodeUnavailable = 52`.

dg-ratiodata:
This should fix #231.

### DIFF
--- a/test/test_ldap_connection.rb
+++ b/test/test_ldap_connection.rb
@@ -47,6 +47,19 @@ class TestLDAPConnection < Test::Unit::TestCase
     end
   end
 
+  def test_result_for_connection_failed_is_set
+    flexmock(TCPSocket).should_receive(:new).and_raise(Errno::ECONNREFUSED)
+
+    ldap_client = Net::LDAP.new(host: '127.0.0.1', port: 12345)
+
+    assert_raise Net::LDAP::ConnectionRefusedError do
+      ldap_client.bind(method: :simple, username: 'asdf', password: 'asdf')
+    end
+
+    assert_equal(ldap_client.get_operation_result.code, 52)
+    assert_equal(ldap_client.get_operation_result.message, 'Unavailable')
+  end
+
   def test_unresponsive_host
     assert_raise Net::LDAP::Error do
       Net::LDAP::Connection.new(:host => 'test.mocked.com', :port => 636)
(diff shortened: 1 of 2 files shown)


test/test_ldap_connection.rb
```diff
@@ -47,6 +47,19 @@ def test_list_of_hosts_with_all_hosts_failure
     end
   end
 
+  def test_result_for_connection_failed_is_set
+    flexmock(TCPSocket).should_receive(:new).and_raise(Errno::ECONNREFUSED)
+
+    ldap_client = Net::LDAP.new(host: '127.0.0.1', port: 12345)
+
+    assert_raise Net::LDAP::ConnectionRefusedError do
```
dg-ratiodata: I'm not not a reqular user of minitest. If there's a better approach to handle errors, I'm open to fix this. `begin; rescue; end` didn't work for that use case.

dg-ratiodata:
@padde Again FYI

jch:
Thanks for this PR. I've been rescuing the errors and assuming the result code is invalid, but this avoids surprising and unexpected behavior. From RFC 4511:

```
      unavailable (52)
         Indicates that the server is shutting down or a subsystem
         necessary to complete the operation is offline.
```

Seems like a good fit.

dg-ratiodata:
Cool. Thanks a lot! Can you a release 0.12.1 with this fix? This would help me getting a system live. :-)

PR #235, opened 2015-11-10
Lazy initialize Net::LDAP::Connection's internal socket

jch:
This PR's goal is to break the dependency between `Net::LDAP::Connection#initialize` and socket initialization. This would fix the [testing code smell](https://github.com/ruby-ldap/ruby-net-ldap/blob/c89d49377601b17225cc76da4431dd0cd3f294e3/test/test_auth_adapter.rb#L5) of stubbing out opening a TCPSocket whenever we initialize a connection object.

* add private #socket and use it for all socket communication
* remove socket initialization from constructor

I'm looking for early feedback as this is a work in progress because many existing tests expect the constructor to initialize the socket. I don't expect this to break the public API since `Net::LDAP::Connection` is an internal class.

cc @satoryu @javanthropus since I recall one of you asking for a better way to test this.

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -13,6 +13,15 @@ class Net::LDAP::Connection #:nodoc:
     yield self if block_given?
   end
 
+  # Allows tests to parameterize what socket class to use
+  def socket_class
+    @socket_class || TCPSocket
+  end
+
+  def socket_class=(socket_class)
+    @socket_class = socket_class
+  end
+
   def prepare_socket(server)
     socket = server[:socket]
     encryption = server[:encryption]
@@ -28,7 +37,7 @@ class Net::LDAP::Connection #:nodoc:
     errors = []
     hosts.each do |host, port|
       begin
-        prepare_socket(server.merge(socket: TCPSocket.new(host, port)))
+        prepare_socket(server.merge(socket: socket_class.new(host, port)))
         return
       rescue Net::LDAP::Error, SocketError, SystemCallError,
              OpenSSL::SSL::SSLError => e
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -6,15 +6,9 @@ class Net::LDAP::Connection #:nodoc:
   LdapVersion = 3
   MaxSaslChallenges = 10
 
-  def initialize(server)
+  def initialize(server = {})
+    @server = server
     @instrumentation_service = server[:instrumentation_service]
-
-    if server[:socket]
-      prepare_socket(server)
-    else
-      server[:hosts] = [[server[:host], server[:port]]] if server[:hosts].nil?
-      open_connection(server)
-    end
 
     yield self if block_given?
   end
@@ -195,7 +189,7 @@ class Net::LDAP::Connection #:nodoc:
   def read(syntax = Net::LDAP::AsnSyntax)
     ber_object =
       instrument "read.net_ldap_connection", :syntax => syntax do |payload|
-        @conn.read_ber(syntax) do |id, content_length|
+        socket.read_ber(syntax) do |id, content_length|
           payload[:object_type_id] = id
           payload[:content_length] = content_length
         end
@@ -225,7 +219,7 @@ class Net::LDAP::Connection #:nodoc:
   def write(request, controls = nil, message_id = next_msgid)
     instrument "write.net_ldap_connection" do |payload|
       packet = [message_id.to_ber, request, controls].compact.to_ber_sequence
-      payload[:content_length] = @conn.write(packet)
+      payload[:content_length] = socket.write(packet)
     end
   end
   private :write
@@ -599,5 +593,19 @@ class Net::LDAP::Connection #:nodoc:
     end
 
     pdu
+  end
+
+  private
+
+  # Returns a Socket like object used internally to communicate with LDAP server
+  #
+  # Typically a TCPSocket, but can be a OpenSSL::SSL::SSLSocket
+  def socket
+    return @conn if defined? @conn
+
+    # First refactoring uses the existing methods open_connection and
+    # prepare_socket to set @conn. Next cleanup would centralize connection
+    # handling here.
+    open_connection(@server)
   end
 end # class Connection
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -595,9 +595,8 @@ class Net::LDAP::Connection #:nodoc:
     pdu
   end
 
-  private
-
-  # Returns a Socket like object used internally to communicate with LDAP server
+  # Internal: Returns a Socket like object used internally to communicate with
+  # LDAP server.
   #
   # Typically a TCPSocket, but can be a OpenSSL::SSL::SSLSocket
   def socket
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -6,6 +6,14 @@ class Net::LDAP::Connection #:nodoc:
   LdapVersion = 3
   MaxSaslChallenges = 10
 
+  # Initialize a connection to an LDAP server
+  #
+  # :server
+  #   :hosts   Array of tuples specifying host, port
+  #   :host    host
+  #   :port    port
+  #   :socket  prepared socket
+  #
   def initialize(server = {})
     @server = server
     @instrumentation_service = server[:instrumentation_service]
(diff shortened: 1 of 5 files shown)


javanthropus:
@jch It might be better to allow the socket class to be set via the initialize method of `Net::LDAP::Connection` along with the rest of the parameters rather than setting it via an additional call to an attribute modifier after initialization.  Doing so would simplify utilizing alternative socket classes generally, not just in tests, while also keeping the overall interface for the `Connection` class relatively simple.

jch:
@javanthropus the Connection class does that a `socket` parameter. However, I don't want to expose this as a public method on `Net::LDAP`. It doesn't make sense for users to think about messing with sockets.

test/test_ldap_connection.rb
```diff
@@ -9,44 +9,57 @@ def capture_stderr
...
+      ["success.host", 636],
+      ["fail.SocketError", 636],
     ]
-    flexmock(TCPSocket).should_receive(:new).ordered.with(*hosts[0]).once.and_raise(SocketError)
-    flexmock(TCPSocket).should_receive(:new).ordered.with(*hosts[1]).once.and_return(nil)
-    flexmock(TCPSocket).should_receive(:new).ordered.never
-    Net::LDAP::Connection.new(:hosts => hosts)
+    connection = Net::LDAP::Connection.new(:hosts => hosts)
+    connection.socket_class = FakeTCPSocket
+    connection.socket
```
javanthropus: @jch Uses such as these 3 lines are what I'm proposing to simplify a bit.  I think it would look better and be easier to use like this:
```ruby
connection = Net::LDAP::Connection.new(:hosts => hosts, :socket_class => FakeTCPSocket)
```
The idea is for this single call to create the connection and arrange for the opening of the socket using the specified socket class when needed all in a single go rather than requiring the `socket` method to be called to actually open the socket.  If the socket class is not specified, it should default to `TCPSocket`.

This is a fairly common way to handle dependency injection as you're trying to enable here.

javanthropus: BTW, with this change in place, you wouldn't need to add the `socket_class` and `socket_class=` attributes to the class definition of `Connection` at all.

lib/net/ldap/connection.rb
```diff
@@ -6,19 +6,30 @@ class Net::LDAP::Connection #:nodoc:
...
+  #
+  # :server
+  #   :hosts   Array of tuples specifying host, port
+  #   :host    host
+  #   :port    port
+  #   :socket  prepared socket
+  #
+  def initialize(server = {})
+    @server = server
     @instrumentation_service = server[:instrumentation_service]
```
javanthropus: If you add the following here, you can simplify overriding the socket class:
```ruby
@socket_class = server.fetch(:socket_class, TCPSocket)
```

lib/net/ldap/connection.rb
```diff
@@ -6,19 +6,30 @@ class Net::LDAP::Connection #:nodoc:
...
 
+  # Allows tests to parameterize what socket class to use
+  def socket_class
+    @socket_class || TCPSocket
+  end
+
+  def socket_class=(socket_class)
+    @socket_class = socket_class
+  end
+
```
javanthropus: These methods could be removed if you allow the socket class to be set in the initializer.

lib/net/ldap/connection.rb
```diff
@@ -34,7 +45,7 @@ def open_connection(server)
     errors = []
     hosts.each do |host, port|
       begin
-        prepare_socket(server.merge(socket: TCPSocket.new(host, port)))
+        prepare_socket(server.merge(socket: socket_class.new(host, port)))
```
javanthropus: Use `@socket_class` here to finish up my suggestion.

test/test_ldap_connection.rb
satoryu: @javanthropus ++  your idea sounds good :+1:

satoryu:
ah, this test for 1.9.3 support failed since `byebug` supports 2.0.0 or later.  
There are two ways, 
1. bye 1.9.3 support
2. bye `byebug` :bug: 

IMO, the option 1 looks better as [Support for Ruby 1.9.3 has been ended](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/).

test/test_ldap_connection.rb
jch: @javanthropus ah, I totally misunderstood! Parameterizing the socket class rather than the socket instance is a good idea. I'll roll with that.

jch:
@satoryu nice investigation. :+1: to dropping support for 1.9 series, but let's do that in a separate PR. Would you be interested in opening one up?

satoryu:
This PR will pass all tests after the master is merged :+1:

jch:
@satoryu thanks for the update. I've been busy with work, so haven't had time to finish this up.

lib/net/ldap/connection.rb
jch: Fixed in 1aab8c9a86d88c378bbc203449341d61d6e7c2f7

jch:
@satoryu @javanthropus this is ready for another pass if either of you could spare some time.

lib/net/ldap/connection.rb
```diff
@@ -652,4 +661,33 @@ def delete(args)
 
     pdu
   end
+
+  # Internal: Returns a Socket like object used internally to communicate with
+  # LDAP server.
+  #
+  # Typically a TCPSocket, but can be a OpenSSL::SSL::SSLSocket
+  def socket
+    return @conn if defined? @conn
```
satoryu: Out of curiosity, I have a question. 
Why is `defined?` used? it looks working as `if @conn`. but there are a little bit difference between the both. 

```ruby
@foo = 1 # <= defined instance variable
puts 'ok' if @foo # => 'ok'
puts 'ok' if defined? @foo # => 'ok'   so the both are same.

@bar = nil # defined but its value is nil
puts 'ok' if @bar # => nothing
puts 'ok' if defined @bar # => 'ok'  so different behavior
```
I believe that this difference doesn't affect on the behavior of `Net::LDAP`.

javanthropus: @satoryu This is likely done to avoid warnings from the interpreter for using uninitialized instance variables.  The `defined?` method is exempted from the warnings as I recall.

lib/net/ldap/connection.rb
```diff
@@ -652,4 +661,33 @@ def delete(args)
...
+
+    @conn
+  end
+
+  private
+
+  # Wrap around Socket.tcp to normalize with other Socket initializers
+  class DefaultSocket
+    def self.new(host, port, socket_opts = {})
+      Socket.tcp(host, port, socket_opts)
```
javanthropus: Looking at a somewhat old definition of `Socket.tcp`, I'm not sure this invocation will work for setting socket options.

http://ruby-doc.org/stdlib-1.9.3/libdoc/socket/rdoc/Socket.html#method-c-tcp

The newer version of that method has an ability to take the socket options, but you need to pass something along to the `local_host` and `local_port` parameters first.

http://ruby-doc.org/stdlib-2.3.0/libdoc/socket/rdoc/Socket.html#method-c-tcp

jch: @javanthropus this is possible because we dropped support for 1.9.x in [0.13.0](https://github.com/ruby-ldap/ruby-net-ldap/blob/master/History.rdoc#netldap-0130)

lib/net/ldap/connection.rb
jch: :+1:

jch:
Thanks for the review!

lib/net/ldap/connection.rb
javanthropus: Check the documentation I linked for version 2.3.0.  Unless I misunderstood something there, you'll need to set the `local_host` and `local_port` parameters to at least `nil` in order to pass `socket_opts`.  Maybe there is some parameter hacking going on behind the scenes though.

javanthropus:
I was working on posting this sooner, but didn't finish it in time...

The more I think about this, the more I think that the library should be able to take a socket factory class that directly produces socket instances.  This would make it possible to provide transports that are independent of hostnames and ports, such as Unix sockets, for instance.  I'm not sure that such functionality would be broadly useful, but the library wouldn't need to be concerned regardless so long as the read/write interfaces of the socket were preserved.

The library should still be able to provide this logic in a backward compatible manner by using an extended version of the `DefaultSocket` class when the socket factory parameter is not provided.  All of the socket creation and configuration logic would move into that class, and a new method named something like `create` would be added to actually return a socket instance to use.

PR #237, opened 2015-11-11
Release 0.12.1

jch:
* Whitespace formatting cleanup {#236}[https://github.com/ruby-ldap/ruby-net-ldap/pull/236]
* Set operation result if LDAP server is not accessible {#232}[https://github.com/ruby-ldap/ruby-net-ldap/pull/232]

cc @dg-ratiodata


dg-ratiodata:
Thanks.

Issue #238, opened 2015-11-23
passing a string as parameter to :encryption causes unexpected error later on


darix:
```
ldap = Net::LDAP.new :host => 'localhost', :port => 389, :encryption => 'start_tls'
```

```
(net-ldap-0.12.1) /lib/net/ldap/connection.rb:111:in `[]': no implicit conversion of Symbol into Integer (TypeError)
        from (net-ldap-0.12.1) /lib/net/ldap/connection.rb:111:in `setup_encryption'
        from (net-ldap-0.12.1) /lib/net/ldap/connection.rb:27:in `prepare_socket'
        from (net-ldap-0.12.1) /lib/net/ldap/connection.rb:37:in `block in open_connection'
        from (net-ldap-0.12.1) /lib/net/ldap/connection.rb:35:in `each'
        from (net-ldap-0.12.1) /lib/net/ldap/connection.rb:35:in `open_connection'
        from (net-ldap-0.12.1) /lib/net/ldap/connection.rb:16:in `initialize'
        from (net-ldap-0.12.1) /lib/net/ldap.rb:1240:in `new'
        from (net-ldap-0.12.1) /lib/net/ldap.rb:1240:in `new_connection'
        from (net-ldap-0.12.1) /lib/net/ldap.rb:1226:in `use_connection'
        from (net-ldap-0.12.1) /lib/net/ldap.rb:752:in `block in search'
        from (net-ldap-0.12.1) /lib/net/ldap/instrumentation.rb:19:in `instrument'
        from (net-ldap-0.12.1) /lib/net/ldap.rb:751:in `search'
        from (net-ldap-0.12.1) /lib/net/ldap.rb:1138:in `search_root_dse'
        from (net-ldap-0.12.1) /lib/net/ldap.rb:1204:in `paged_searches_supported?'
        from (net-ldap-0.12.1) /lib/net/ldap.rb:744:in `search'
        from ldaptest.rb:11:in `<main>'
```

This should probably raise a an error in ldap.rb  def encryption

darix:
seems fixed in 0.12.1

satoryu:
@darix thank you for your bug report. This failure is caused because `Net::LDAP#encrypt` cannot handle encryption method given as String. I'll fix it later.

satoryu:
This can be closed. @jch

jch:
@satoryu good call

PR #239, opened 2015-11-25
Net::LDAP#encryption accepts string

satoryu:
As #238 reported, if `Net::LDAP.new` is given encryption type as `String` instead of `Symbol`, it breaks in establishing a connection. 

This pull request provides a solution: allowing to give encryption type as `String`. 

CC: @darix

### DIFF
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -590,6 +590,7 @@ class Net::LDAP
   # This method is deprecated. 
   #
   def encryption(args)
+    warn "Deprecation warning: please give :encryption option as a Hash to Net::LDAP.new"
     return if args.nil?
     return @encryption = args if args.is_a? Hash
 
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -590,6 +590,7 @@ class Net::LDAP
   #   }
   def encryption(args)
     return if args.nil?
+    return @encryption = args if args.is_a? Hash
 
     case method = args.to_sym
     when :simple_tls, :start_tls
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -461,11 +461,52 @@ class Net::LDAP
   #   call to #search, that value will override any treebase value you give
   #   here.
   # * :encryption => specifies the encryption to be used in communicating
-  #   with the LDAP server. The value is either a Hash containing additional
-  #   parameters, or the Symbol :simple_tls, which is equivalent to
-  #   specifying the Hash {:method => :simple_tls}. There is a fairly large
-  #   range of potential values that may be given for this parameter. See
-  #   #encryption for details.
+  #   with the LDAP server. The value must be a Hash containing additional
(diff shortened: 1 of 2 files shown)


jch:
@satoryu thanks for starting this. I think your code is correct, but I'd rather we address underlying problem: we should have only one way of specifying encryption settings. I always thought it strange to have a method instead of a setter / constructor argument for setting encryption args. `initialize` already calls `encryption`, but I'd prefer it didn't have any special logic and always expected a hash.

This is where I'd like to get to:

```ruby
class Net::LDAP
  # encryption: Hash
  #   method: - :simple_tls (let's standardize on one key, not multiple aliases)
  #   options: - Hash of options for that method
  def initialize(args)
    @encryption = args[:encryption]
  end
end
```

I imagine we'd need to:

* deprecate `#encryption`
* deprecate multiple method names `start_tls`, `simple_tls` in favor of one
* update docs
* update tests

darix:
imho the only weird part about this method is that it doesn't use the assigning signature

```ruby
def encryption=(args)
...
end
```

was the first method i was looking for. but other than that it is a nice way to set the setting. and it matches the function <=> option mapping you see with the other params for the initializer.

jch:
@darix following the setter convention would be less surprising, but the problem is setting `@encryption` only makes sense before a connection is opened. If a caller sets encryption after open, all the behavior is unexpected/undefined. Should it do nothing? That'd be bad because the caller expects the connection to be encrypted. Should we close and reopen the existing connection? That'd be hard to debug too.

Thanks for that initial bug report to spark this discussion :+1:

darix:
On 2015-11-30 09:56:06 -0800, Jerry Cheung wrote:
> @darix following the setter convention would be less surprising, but
> the problem is setting `@encryption` only makes sense before a
> connection is opened. If a caller sets encryption after open, all the
> behavior is unexpected/undefined. Should it do nothing? That'd be bad
> because the caller expects the connection to be encrypted. Should we
> close and reopen the existing connection? That'd be hard to debug too.

I do not see how that is much different from calling

```ruby
ldap_object.encryption(somearg)
```

once the connection is established.

We would have 2 options in that case.

a) reject the setting with an error Net::LDAP::OpenConnectionError
b) close the current connection and reopen it with the new settings.

the second option might have unwanted consequences like incomplete
operations, especially relevant for write operations. so the first
option might be better.

satoryu:
@jch I agree with your idea about deprecating `encryption` and two options `start_tls` and `simple_tls` and only supporting Hash option for specifying encryption type. As this change would break some user codes which expects these, the minor version should be bumped up. 

@darix I prefer option a) and showing deprecation massage when calling `encryption` method.

jch:
@satoryu any interest in PR-ing the ideas listed above? If you're busy, I'll try to carve out some time during the holidays.

satoryu:
@jch sorry for late response. I'd like to keep doing PRing in this week end.

jch:
@satoryu happy new years, and thanks for following up!

PR #240, opened 2015-12-01
Drop support for ruby 1.9.3

satoryu:
As we discussed in #235, now is time to say bye to ruby 1.9.3. 

haha, net-ldap is already ready for support 2.3.0 :tada:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: ruby
 rvm:
-  - 1.9.3
   - 2.0.0
   - 2.1
   - 2.2
   # optional
+  - 2.3.0-preview1
   - ruby-head
   - jruby-19mode
   - jruby-head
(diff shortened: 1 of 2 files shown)


jch:
:clap:

.travis.yml
```diff
@@ -1,10 +1,10 @@
 language: ruby
 rvm:
-  - 1.9.3
   - 2.0.0
   - 2.1
   - 2.2
   # optional
+  - 2.3.0-preview1
   - ruby-head
   - jruby-19mode
```
jch: Ah, shoot. I merged this too soon. We should get rid of this also.

.travis.yml
```diff
@@ -21,6 +21,7 @@ script: bundle exec rake ci
 
 matrix:
   allow_failures:
+    - rvm: 2.3.0-preview1
```
satoryu: should we still get rid of this even if it is allowed to fail ?

jch: I think we can skip the previews since we already include Ruby-head

PR #243, opened 2015-12-17
Set a connect_timeout for the creation of a socket

astratto:
Currently when an LDAP server is not reachable (for instance, packets are dropped), users of this library end up waiting for a quite long period of time. This is very rarely acceptable and can easily lead to a chain of failures with several services timing out before the socket creation actually fails.
 
This patch uses Socket.tcp instead of TCPSocket.new and provides a default connect_timeout of 1 second.

_Notes:_
- I couldn't find a way in the integration tests to write a proper test for this. As far as I can tell, there are no tests at that level to validate connection failures, etc... Am I right?
- I set the timeout to 1 sec, but probably a better way could be setting it to a higher value and provide an easier way to customise it?

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -3,6 +3,9 @@
 class Net::LDAP::Connection #:nodoc:
   include Net::LDAP::Instrumentation
 
+  # Seconds before failing for socket connect timeout
+  DefaultConnectTimeout = 5
+
   LdapVersion = 3
   MaxSaslChallenges = 10
 
@@ -31,10 +34,14 @@ class Net::LDAP::Connection #:nodoc:
     hosts = server[:hosts]
     encryption = server[:encryption]
 
+    socket_opts = {
+      connect_timeout: server[:connect_timeout] || DefaultConnectTimeout
+    }
+
     errors = []
     hosts.each do |host, port|
       begin
-        prepare_socket(server.merge(socket: TCPSocket.new(host, port)))
+        prepare_socket(server.merge(socket: Socket.tcp(host, port, socket_opts)))
         return
       rescue Net::LDAP::Error, SocketError, SystemCallError,
              OpenSSL::SSL::SSLError => e
--- a/script/install-openldap
+++ b/script/install-openldap
@@ -109,4 +109,7 @@ chgrp ssl-cert /etc/ssl/private/ldap01_slapd_key.pem
 chmod g+r /etc/ssl/private/ldap01_slapd_key.pem
 chmod o-r /etc/ssl/private/ldap01_slapd_key.pem
 
+# Drop packets on a secondary port used to specific timeout tests
+iptables -A OUTPUT -p tcp -j DROP --dport 8389
+
 service slapd restart
--- a/test/test_auth_adapter.rb
+++ b/test/test_auth_adapter.rb
@@ -2,7 +2,8 @@ require 'test_helper'
 
 class TestAuthAdapter < Test::Unit::TestCase
   def test_undefined_auth_adapter
-    flexmock(TCPSocket).should_receive(:new).ordered.with('ldap.example.com', 379).once.and_return(nil)
+    flexmock(Socket).should_receive(:tcp).ordered.with('ldap.example.com', 379, { connect_timeout: 5 }).once.and_return(nil)
+
     conn = Net::LDAP::Connection.new(host: 'ldap.example.com', port: 379)
     assert_raise Net::LDAP::AuthMethodUnsupportedError, "Unsupported auth method (foo)" do
       conn.bind(method: :foo)
--- a/lib/net/ber.rb
+++ b/lib/net/ber.rb
@@ -293,24 +293,24 @@ end
 
 ##
 # A String object with a BER identifier attached.
-# 
+#
 class Net::BER::BerIdentifiedString < String
   attr_accessor :ber_identifier
 
   # The binary data provided when parsing the result of the LDAP search
   # has the encoding 'ASCII-8BIT' (which is basically 'BINARY', or 'unknown').
-  # 
+  #
   # This is the kind of a backtrace showing how the binary `data` comes to
   # BerIdentifiedString.new(data):
   #
   #  @conn.read_ber(syntax)
   #     -> StringIO.new(self).read_ber(syntax), i.e. included from module
-  #     -> Net::BER::BERParser.read_ber(syntax) 
+  #     -> Net::BER::BERParser.read_ber(syntax)
   #        -> (private)Net::BER::BERParser.parse_ber_object(syntax, id, data)
-  # 
+  #
   # In the `#parse_ber_object` method `data`, according to its OID, is being
   # 'casted' to one of the Net::BER:BerIdentifiedXXX classes.
-  # 
+  #
   # As we are using LDAP v3 we can safely assume that the data is encoded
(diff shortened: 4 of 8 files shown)


jch:
@astratto interesting. It looks like `Socket.tcp` gained some options in 2.x that weren't available in 1.9.x. Since we've dropped support for 1.9, this is now possible.

> As far as I can tell, there are no tests at that level to validate connection failures, etc... Am I right?

Yes. I don't know how to simulate a hanging connection. If you have ideas, I'd love to hear them.

> I set the timeout to 1 sec, but probably a better way could be setting it to a higher value and provide an easier way to customise it?

Agree this should be parameterized. `:connect_timeout` seems like a good option name. A unit test we could write would be to verify that Socket.tcp is called with the parameter.

Thanks for opening this pull!

astratto:
> As far as I can tell, there are no tests at that level to validate connection failures, etc... Am I right?
> Yes. I don't know how to simulate a hanging connection. If you have ideas, I'd love to hear them.

I actually already have tests for our projects to trigger that situation (hence the PR :smile:), I'm just not sure how to do that with the integration tests here.

Basic idea is to use a firewall to drop packets.
A simple ```iptables -A OUTPUT -j DROP -p tcp --dport $LDAP_PORT``` would do the trick in this case.

>> I set the timeout to 1 sec, but probably a better way could be setting it to a higher value and provide an easier way to customise it?

> Agree this should be parameterized. :connect_timeout seems like a good option name. A unit test we could write would be to verify that Socket.tcp is called with the parameter.

Yes, I just wanted to understand your point of view on the general PR.
I'll try to work on that tomorrow, but feel free to jump on it if you have time.

> Thanks for opening this pull!

Thanks for this gem :wink:

jch:
The entry point for integration tests on travis is in https://github.com/ruby-ldap/ruby-net-ldap/blob/master/.travis.yml#L15-L17. This then runs `script/install-openldap`. I'm not sure whether travis supports iptables, but we could run your rule for the duration of one test and then clean up after that.

astratto:
Hey @jch, I've just pushed a patch to expose :connect_timeout and :timeout.

I think it's worth trying to add some integration tests for connection errors in general, should we address that in another patchset?

lib/net/ldap/connection.rb
```diff
@@ -31,10 +36,15 @@ def open_connection(server)
     hosts = server[:hosts]
     encryption = server[:encryption]
 
+    socket_opts = {
+      connect_timeout: server[:connect_timeout] || DefaultConnectTimeout,
+      timeout: server[:timeout] || DefaultTimeout
```
satoryu: According to [the doc](http://docs.ruby-lang.org/en/2.2.0/Socket.html#method-c-tcp), `Socket.tcp` deals with only `connect_timeout`. Does `timeout` make sense?

astratto: Interesting, it was there in 2.0.0... http://docs.ruby-lang.org/en/2.0.0/Socket.html#method-c-tcp
I'm gonna remove it :smile:

astratto: Gone :wink:

satoryu: well-done! :+1:

jch:
> I think it's worth trying to add some integration tests for connection errors in general, should we address that in another patchset?

:+1: that's be amazing!

lib/net/ldap/connection.rb
```diff
@@ -3,6 +3,9 @@
 class Net::LDAP::Connection #:nodoc:
   include Net::LDAP::Instrumentation
 
+  # Seconds before failing for socket connect timeout
+  DefaultConnectTimeout = 1
```
jch: Could we make this value a bit more generous so it doesn't surprise any users with slow connections?

test/test_auth_adapter.rb
```diff
@@ -2,7 +2,11 @@
 
 class TestAuthAdapter < Test::Unit::TestCase
   def test_undefined_auth_adapter
-    flexmock(TCPSocket).should_receive(:new).ordered.with('ldap.example.com', 379).once.and_return(nil)
+    flexmock(Socket).should_receive(:tcp)
+                    .ordered
+                    .with('ldap.example.com', 379, SocketHelpers.default_socket_opts)
+                    .once.and_return(nil)
+
```
jch: /cc https://github.com/ruby-ldap/ruby-net-ldap/pull/235

test/test_helper.rb
```diff
@@ -64,3 +64,12 @@ def setup
...
 end
+
+# Provide information about socket creation
+class SocketHelpers
+  class << self
+    def default_socket_opts
+      { connect_timeout: Net::LDAP::Connection::DefaultConnectTimeout }
+    end
+  end
+end
```
jch: This seems overkill. Could you inline this to the tests? My testing preferences are in https://github.com/jch/styleguide/blob/master/rails.md#tests

jch: Also, it'd be good to not depend on the implementation in the test. So avoid referencing Net::LDAP::Connection::DefaultConnectTimeout in the test, and use the actual value instead. If that changes, then it's easy to see where the failures are and replace them.

astratto: Yes, I see your point. Changed.

test/test_auth_adapter.rb
astratto: Just to chime in here: that would definitely simplify a lot of tests :+1:

lib/net/ldap/connection.rb
astratto: Sure, it's 5 seconds now.
I doubt more than that would be of any use, but let me know if you want to bump it further.

astratto:
@jch I'd like to decouple this PR with the integration tests.
I've just created https://github.com/ruby-ldap/ruby-net-ldap/issues/244 to track that.

astratto:
Okay, it was actuallystraight forward given that the box is used by Travis has iptables available.
There's one integration test for this now.

jch:
@astratto looks like there's a travis error. Could we re-run it somehow?

satoryu:
@astratto this failure seems like https://github.com/spotify/rspec-dns/pull/26 . 

so will be fixed by adding the following lines to `.travis.yml`. 
```yaml
before_install:
  - gem update bundler
```

satoryu:
@jch @astratto I sent another PR #245 to fix this failure.

astratto:
@jch I've just rebased and now it's all green.

Thanks @satoryu

script/install-openldap
```diff
@@ -109,4 +109,7 @@ chgrp ssl-cert /etc/ssl/private/ldap01_slapd_key.pem
 chmod g+r /etc/ssl/private/ldap01_slapd_key.pem
 chmod o-r /etc/ssl/private/ldap01_slapd_key.pem
 
+# Drop packets on a secondary port used to specific timeout tests
```
jch: :+1: helpful comment

jch:
Looks great. I especially appreciate you taking the time to add an integration test. :tada:

astratto:
Thanks @jch 

It'd be nice to add more of those tests, let's see if I manage to find some spare time in the next few weeks.

jch:
Keep on rocking :metal:

satoryu:
Thank you for your great works :tada:

dblessing:
I can't get the `connect_timeout` to actually time out as expected. See #247. Is this a bug or am I using it wrong?

PR #245, opened 2016-01-05
Update bundler before installing gems with bundler

satoryu:
As I mentioned, #243 and the current HEAD of the master branch are facing with Travis-CI's bundler issue [[FYI]](https://github.com/spotify/rspec-dns/pull/26). 

This pull request provides a solution which updates bundler before bundling.


jch:
Thank you!

PR #249, opened 2016-01-08
Rubocop style fixes

jch:
This PR fixes several:

* multi line blocks with `do` `end` instead of braces
* trailing underscore in multi assignment

I've added a few explicit exceptions for the project, and regenerated the `.rubocop_todo.yml` by running:

```
$ rubocop --auto-gen
```

Feel free to pick up and try to clean any of these up. If you have questions about whether we want to follow a certain style, open an issue and we can discuss.

cc @mynameisrufus since you added the gem originally
cc @satoryu


mynameisrufus:
@jch sure thing I will start picking them (the uncontroversial ones) off.

jch:
@mynameisrufus thanks! If you could open them in separate PR's, one per style adjustment, that would make it easier to review and discuss.

mynameisrufus:
Just to jog your memory, at the time it was added it was removed from the build. Maybe consider adding it in again https://github.com/ruby-ldap/ruby-net-ldap/blob/master/Rakefile#L18

PR #252, opened 2016-01-12
fix space inside brackets

mynameisrufus:


jch:
Interesting, I always assumed since we did a space between braces, we'd do something similar with brackets. I'm fine with this though since it would keep us closer to rubocop defaults.

PR #254, opened 2016-01-12
fix deprecated hash methods

mynameisrufus:


lib/net/snmp.rb
```diff
@@ -191,7 +191,7 @@ def pdu_type= t
     end
 
     def error_status= es
-      unless ErrorStatusCodes.has_key?(es)
+      unless ErrorStatusCodes.key?(es)
```
jch: Today I learned ;)

satoryu: me too, thanks!

PR #256, opened 2016-01-12
fix trailing commas

mynameisrufus:


jch:
Could we add an exception for this rule? I actually prefer trailing commas in hashes because keeps diffs line oriented. http://mislav.net/2014/02/hidden-documentation/ explains:

> Corollary of avoiding unrelated changes: stick to a line-based coding style that allows you to append, edit or remove values from lists without changing adjacent lines. Well, always think about the person who’s going to git blame this. In the JavaScript example, if you were the one who added a committed the value "baz", you don’t want your name to show up when somebody blames the line that added "bar", since the two variables might be unrelated.

mynameisrufus:
Golang style. I can't disagree with this. PR updated with enforced commas.

jch:
Ha! Of course there's another rule that would enforce this. Thanks, merging.

Issue #250, opened 2016-01-11
Encryption Bug


tonobo:
Hey, 0.12.1 is incompatible with 0.13.0.  I get the following trace, with encryption enabled.

"no implicit conversion of Symbol into Integer"

net-ldap (0.13.0) lib/net/ldap/connection.rb:118:in `[]'
net-ldap (0.13.0) lib/net/ldap/connection.rb:118:in`setup_encryption'
net-ldap (0.13.0) lib/net/ldap/connection.rb:30:in `prepare_socket'
net-ldap (0.13.0) lib/net/ldap/connection.rb:44:in`block in open_connection'
net-ldap (0.13.0) lib/net/ldap/connection.rb:42:in `each'
net-ldap (0.13.0) lib/net/ldap/connection.rb:42:in`open_connection'
net-ldap (0.13.0) lib/net/ldap/connection.rb:19:in `initialize'
net-ldap (0.13.0) lib/net/ldap.rb:1254:in`new'
net-ldap (0.13.0) lib/net/ldap.rb:1254:in `new_connection'
net-ldap (0.13.0) lib/net/ldap.rb:849:in`block in bind'
net-ldap (0.13.0) lib/net/ldap/instrumentation.rb:19:in `instrument'
net-ldap (0.13.0) lib/net/ldap.rb:843:in`bind'

jch:
@timmyArch could you include the snippet of code you're calling with?

timmyArch:
@jch this will do the trick ;)

ldap = Net::LDAP.new(
      :host => LDAP_HOST,
      :port => LDAP_PORT,
      :base => LDAP_BASE,
      :encryption => :simple_tls,
      :auth => { :method => :simple, :username => user_dn, :password => password }
    )
ldap.bind

jch:
@timmyArch The constructor way of initializing encryption requires a hash (https://github.com/ruby-ldap/ruby-net-ldap/pull/239). This is different than old way of setting it through the `#encryption` method, but it does standardize the param. Your example should work with:

```ruby
ldap = Net::LDAP.new(:encryption => { :simple_tls }, ...)
ldap.bind
```

@satoryu I noticed the "Documentation" link on https://rubygems.org/gems/net-ldap/versions/0.13.0 links to a really old version of net-ldap (0.0.4). I tried to file a bug on rubydoc.info, but got a 500. Any ideas?

timmyArch:
[...]

PR #257, opened 2016-01-13
Raises errors when giving encryption as non-Hash object

satoryu:
In #250 , `Net::LDAP` of the latest version 0.13.0 accepts encryption options on its initializer. It causes an error when creating a connection as reported. 

This PR provides a validation which check whether a given `encryption` option is a Hash or not.

### DIFF
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -540,6 +540,9 @@ class Net::LDAP
     @base = args[:base] || DefaultTreebase
     @force_no_page = args[:force_no_page] || DefaultForceNoPage
     @encryption = args[:encryption] # may be nil
+    if !@encryption.nil? and !@encryption.is_a?(Hash)
+      raise ArgumentError, "encryption must be given as Hash"
+    end
     @connect_timeout = args[:connect_timeout]
 
     if pr = @auth[:password] and pr.respond_to?(:call)
--- a/test/test_ldap.rb
+++ b/test/test_ldap.rb
@@ -91,4 +91,10 @@ class TestLDAPInstrumentation < Test::Unit::TestCase
 
     assert_equal enc[:method], :start_tls
   end
+
+  def test_initialize
+    assert_raise ArgumentError, "encryption must be given as Hash" do
+      Net::LDAP.new encryption: [ :simple_tls ]
+    end
+  end
 end


timmyArch:
:+1:  it would be nice to validate all given parameters from constructor.

lib/net/ldap.rb
```diff
@@ -540,6 +540,9 @@ def initialize(args = {})
     @base = args[:base] || DefaultTreebase
     @force_no_page = args[:force_no_page] || DefaultForceNoPage
     @encryption = args[:encryption] # may be nil
+    if !@encryption.nil? and !@encryption.is_a?(Hash)
```
jch: Can we switch this to `@encryption && !@encryption.is_a?(Hash)`? This makes the nil check implicit, and the `&&` operator is higher precedence than `and` (doesn't matter in this case, but I prefer it when used in a conditional)

test/test_ldap.rb
```diff
@@ -91,4 +91,10 @@ def test_encryption
 
     assert_equal enc[:method], :start_tls
   end
+
+  def test_initialize
```
jch: Good test, but could we rename this to something more descriptive? Perhaps `test_initialize_requires_hash_encryption`?

lib/net/ldap.rb
mynameisrufus: These operators still hurt my brain, if we replace the `and` operators in the this codebase to `&&` all the tests fail.

http://devblog.avdi.org/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/

jch: Alternatively, we could also use a single type check with `@encryption.respond_to?(:[])`. I don't have a strong preference here.

@mynameisrufus why would `&&` fail here?

mynameisrufus: Not here specifically, but when trying to replace `and` with `&&` throughout the entire codebase (I tried it) all the test fail :)

jch: Thanks for clarifying. I wasn't trying to make so bold a claim ;)

satoryu: @jch @mynameisrufus Thank you for pointing out this condition. made my understanding about these operators clear :+1:

PR #261, opened 2016-01-23
Update Docs: Net::LDAP now requires ruby >= 2

jesjos:
Ran into this while installing the gem on JRuby 1.7.24


satoryu:
Thanks! 

@jch could you please merge this PR?

jch:
@jesjos thanks for reporting!

PR #262, opened 2016-01-23
SSL hostname verification support

tarcieri:
The current `Net::LDAP::Connection::wrap_with_ssl` verifies the certification paths of SSL certificates correctly, but it does not verify hostnames. This means any service with a certificate issued by the same CA as the LDAP server can impersonate the LDAP server. (Note: this is probably CVE worthy. Let me know if you'd like help getting a CVE assigned)

This PR is more of a strawman implementation than anything to get the ball rolling towards a solution. I think it could definitely use some tests to ensure hostname verification works! But to do that we might want to talk about how to issue all of the certificates needed in the tests.

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -36,7 +36,7 @@ class Net::LDAP::Connection #:nodoc:
     encryption = server[:encryption]
 
     @conn = socket
-    setup_encryption encryption if encryption
+    setup_encryption({ verify_host: server[:connected_host] }.merge(encryption)) if encryption
   end
 
   def open_connection(server)
@@ -50,7 +50,7 @@ class Net::LDAP::Connection #:nodoc:
     errors = []
     hosts.each do |host, port|
       begin
-        prepare_socket(server.merge(socket: @socket_class.new(host, port, socket_opts)))
+        prepare_socket(server.merge(socket: @socket_class.new(host, port, socket_opts), connected_host: host))
         return
       rescue Net::LDAP::Error, SocketError, SystemCallError,
              OpenSSL::SSL::SSLError => e
@@ -87,6 +87,13 @@ class Net::LDAP::Connection #:nodoc:
 
     conn = OpenSSL::SSL::SSLSocket.new(io, ctx)
     conn.connect
+
+    if tls_options[:verify_host]
+      # This raises OpenSSL::SSL::SSLError if hostname verification fails
+      conn.post_connection_check(tls_options[:verify_host])
+    else
+      warn "not verifying SSL hostname of LDAP server"
+    end
 
     # Doesn't work:
     # conn.sync_close = true


tarcieri:
I tried to write this so it's possible to pass `tls_options[:verify_host] = false` although it is not the default. I am guessing there are people who do not have the correct subjectAltNames for the SSL certificates on their LDAP servers, and this patch if done correctly will make that break.

I made the code `warn` every time a connection is made without verifying hostnames, so hopefully this provides a nice tradeoff between being able to turn the security off while still being sufficiently annoying if you do.

lib/net/ldap/connection.rb
```diff
@@ -88,6 +88,13 @@ def self.wrap_with_ssl(io, tls_options = {})
     conn = OpenSSL::SSL::SSLSocket.new(io, ctx)
     conn.connect
 
+    if tls_options[:verify_host]
+      # This raises OpenSSL::SSL::SSLError if hostname verification fails
+      conn.post_connection_check(tls_options[:verify_host])
```
tarcieri: A previously undocumented (until I documented it) method it is very important to call if you want your SSL connections to be secure!

JPvRiel:
So this is a second pull request to address this same issue https://github.com/ruby-ldap/ruby-net-ldap/issues/258. My initial hack was pull https://github.com/ruby-ldap/ruby-net-ldap/pull/259.

This pull request looks neater and more funcitonal than https://github.com/ruby-ldap/ruby-net-ldap/pull/259 in terms of parsing the hostname info through the function calls and doing the validation according to the `:verify_host` option. However, arguably, we should aim for secure by default (which https://github.com/ruby-ldap/ruby-net-ldap/pull/259 did). May I suggest the following logic:
* when using LDAPS, `verify_host` (and validate the cert chain) by default (and in newer versions of ruby VERIFY_PEER is at least also a default)
* have a `:allow_host_missmatch` negative option instead of `:verify_host` for people who want or need to disable it
* also, don't bother to try verify (match) the hostname if users have forced ` :verify_mode => OpenSSL::SSL::VERIFY_NONE` and the issue that warning. This will also imply it's backward  compatible and insecure by default when run with older versions of ruby that defaulted to VERIFY_NONE.

Users could force old ruby behaviour with, but arguably they should just use plain LDAP / 389 instead then.
```
:tls_options => { :verify_mode => OpenSSL::SSL::VERIFY_NONE }
```

lib/net/ldap/connection.rb
```diff
@@ -36,7 +36,7 @@ def prepare_socket(server)
     encryption = server[:encryption]
 
     @conn = socket
-    setup_encryption encryption if encryption
+    setup_encryption({ verify_host: server[:connected_host] }.merge(encryption)) if encryption
```
tarcieri: Note that this does verification by default unless explicitly disabled

jch:
@tarcieri @JPvRiel thanks for starting these. I've been busy with work, so I haven't had time to review either of these in depth. @satoryu any initial thoughts on these?

lib/net/ldap/connection.rb
JPvRiel: @tarcieri, sorry I'm not really much of a ruby dev, so I may misunderstand a bit. I assume any non nil value in a conditional returns false... so users must explicitly enable verification in config with `:encryption = { :method => :simple_tls, :tls_options => { :verify_host => true } }`?

If I understand correctly, `:encryption => { :method => :simple_tls, :tls_options => { :verify_mode => OpenSSL::SSL::VERIFY_PEER, :tls_options[:verify_host] => false} }` should imply checking that the cert is at least signed by a trusted CA, but allowing for a mismatch?

What happens if the user has simply put `{ :method => :simple_tls, :tls_options => { :verify_mode => OpenSSL::SSL::VERIFY_PEER`}`?

```
+    if tls_options[:verify_host]
 +      # This raises OpenSSL::SSL::SSLError if hostname verification fails
 +      conn.post_connection_check(tls_options[:verify_host])
```
The match function will not run if tls_options[:verify_host] is nil? Hence I thought that your pull doesn't do hostname validation unless the user explicitly configures it.

tarcieri: No, this is passing the hostname through as `verify_host`, *unless* the user has set `verify_host` to something else.

It's a bit convoluted, but it's on by default, and enabled the user to opt-out.

I would agree having a separate user targeted flag would make sense, but you don't seem to be understanding how this code is working.

JPvRiel: @tarcieri, thanks, got it now. I cloned your fork and tested `:verify_host => false` (with valid variables set for my env). It bumps into an error.

Can you provide some config examples?

```ruby
  conn = Net::LDAP.new :host => hostname,
                       :port => 636,
                       :base => base,
                       :encryption => { :method => :simple_tls,
                                        :tls_options => { :verify_mode => OpenSSL::SSL::VERIFY_PEER,
                                                          :verify_host => false,
                                                          :ca_file => ca_file } },
                       :auth => { :username => login,
                                  :password => pass,
                                  :method => :simple }
  conn.bind 
/usr/lib/ruby/2.1.0/openssl/ssl.rb:87:in `block in set_params': undefined method `verify_host=' for #<OpenSSL::SSL::SSLContext:0x000000016f4130> (NoMethodError)
	from /usr/lib/ruby/2.1.0/openssl/ssl.rb:87:in `each'
	from /usr/lib/ruby/2.1.0/openssl/ssl.rb:87:in `set_params'
	from /home/a211278l/Own Tech Examples/Ruby/LDAPS/ruby-net-ldap-fix2/lib/net/ldap/connection.rb:86:in `wrap_with_ssl'

```

tarcieri:
@JPvRiel your PR is fine. I'd say let's go with that

JPvRiel:
@tarcieri I thought your approach was technically neater by passing the host address info through the function calls to `wrap_with_ssl` which is where one would logically expect to perform SSL checks. My fix was just hacked into the most convenient place.

Regardless, I also saw notes that the connection handling stuff might be refactored...

PR #264, opened 2016-01-27
Normalize the encryption parameter passed to the LDAP constructor

ryanshow:
This PR restores API functionality that was broken on 0.13, the encryption parameter can again be specified as a symbol and will be normalized to the expected hash.

This is in reference to ruby-ldap/ruby-net-ldap#250

### DIFF
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -539,7 +539,7 @@ class Net::LDAP
     @auth = args[:auth] || DefaultAuth
     @base = args[:base] || DefaultTreebase
     @force_no_page = args[:force_no_page] || DefaultForceNoPage
-    @encryption = args[:encryption] # may be nil
+    @encryption = normalize_encryption(args[:encryption]) # may be nil
     @connect_timeout = args[:connect_timeout]
 
     if pr = @auth[:password] and pr.respond_to?(:call)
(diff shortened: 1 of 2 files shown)


lib/net/ldap.rb
```diff
@@ -1323,4 +1317,17 @@ def new_connection
...
+  # convenience symbols into recognizable hashes
+  def normalize_encryption(args)
+    return if args.nil?
+    return args if args.is_a? Hash
+
+    case method = args.to_sym
+    when :simple_tls, :start_tls
+      { :method => method, :tls_options => {} }
+    end
+  end
```
satoryu: Looks good! 
Could you make `normalize_encryption` do what #257 does?

lib/net/ldap.rb
```diff
@@ -609,13 +609,7 @@ def authenticate(username, password)
   def encryption(args)
     warn "Deprecation warning: please give :encryption option as a Hash to Net::LDAP.new"
     return if args.nil?
```
jch: You already moved this line to your `normalize_encryption` method.

ryanshow: Yeah, I debated whether to include that line in there or not. I'm trying to mimic the exact logic that currently exists, which is "if nil is passed to the encryption method, don't set the encryption". If I omit the line, It will set the encryption to nil.

I'm fine with changing it if you feel it is the desired behavior to have the method set the encryption to nil.

lib/net/ldap.rb
ryanshow: Sorry, what exactly would you like me to do? The method should take 1 of 4 types of arguments (nil, string, symbol, or hash) and return either nil or a hash.

lib/net/ldap.rb
jch: That's a great point. :+1: to leaving this. Especially since it's a deprecated method.

lib/net/ldap.rb
satoryu: ah, sorry, I misunderstood. the purpose of this method is to normalize (I was thinking of validating a given `encryption` option when I saw your PR). 
Anyway, please forget about my comment :bow:

PR #268, opened 2016-04-04
Added private recursive_delete as alternative to DELETE_TREE

duffyjp:
`delete_tree` checks if you LDAP server supports the DELETE_TREE control code and then either uses it, or recursively deletes instead-- to the same effect.

### DIFF
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -1170,14 +1170,22 @@ class Net::LDAP
   # entries. This method sends an extra control code to tell the LDAP server
   # to do a tree delete. ('1.2.840.113556.1.4.805')
   #
+  # If the LDAP server does not support the DELETE_TREE control code, subordinate
+  # entries are deleted recursively instead.
+  #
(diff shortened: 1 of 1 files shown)


duffyjp:
I tracked down the source code for "ldapdelete" and it looks like that always does a recursive delete, rather than using the control code.  I see no reason not to use DELETE_TREE if it's available though.

https://opensource.apple.com/source/OpenLDAP/OpenLDAP-208.5/OpenLDAP/clients/tools/ldapdelete.c


Here's an interesting chunk.  If you pass ```-r``` it does a recursive delete, if not it just tries to delete the entity and fails if it has children.  Like ruby-net-ldap's ```delete```.  

```c

	/* If prune is on, remove a whole subtree.  Delete the children of the
	 * DN recursively, then the DN requested.
	 */
	if ( prune ) {
retry:;
		deletechildren( ld, dn, subentries );
	}

	rc = ldap_delete_ext( ld, dn, NULL, NULL, &id );
```

PR #271, opened 2016-05-18
Send DN and newPassword with password_modify request

thriqon:
This change allows an admin to overwrite the password of a user a) without specifying the old one and b) setting the new one.

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -593,11 +593,11 @@ class Net::LDAP::Connection #:nodoc:
 
     ext_seq = [Net::LDAP::PasswdModifyOid.to_ber_contextspecific(0)]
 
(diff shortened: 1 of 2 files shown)


sonOfRa:
This would be very useful for my setup, is it possible to get this merged soon-ish?

jirutka:
@mtodd, @tmaher: Could you please merge this and release a new version? I spent half an hour trying to figure out why resetting password of an user doesn’t work (LDAP returns very confusing error message) until I found that it’s a bug in this gem.

mtodd:
This is now included in the v0.16.1 release and has been pushed up to RubyGems: https://rubygems.org/gems/net-ldap/versions/0.16.1

Thanks for putting together this quality PR @thriqon, and apologies for the slow turnaround on getting this merged and released. Your contribution is much appreciated!

PR #273, opened 2016-06-15
Respect connect_timeout when establishing SSL connections

tmm1:
cc #243

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -31,26 +31,27 @@ class Net::LDAP::Connection #:nodoc:
     @socket_class = socket_class
   end
 
-  def prepare_socket(server)
+  def prepare_socket(server, timeout=nil)
     socket = server[:socket]
     encryption = server[:encryption]
 
     @conn = socket
-    setup_encryption encryption if encryption
+    setup_encryption(encryption, timeout) if encryption
   end
 
   def open_connection(server)
     hosts = server[:hosts]
     encryption = server[:encryption]
 
+    timeout = server[:connect_timeout] || DefaultConnectTimeout
     socket_opts = {
-      connect_timeout: server[:connect_timeout] || DefaultConnectTimeout,
+      connect_timeout: timeout,
     }
 
     errors = []
     hosts.each do |host, port|
       begin
-        prepare_socket(server.merge(socket: @socket_class.new(host, port, socket_opts)))
+        prepare_socket(server.merge(socket: @socket_class.new(host, port, socket_opts)), timeout)
         return
       rescue Net::LDAP::Error, SocketError, SystemCallError,
              OpenSSL::SSL::SSLError => e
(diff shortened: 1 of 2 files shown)


lib/net/ldap/connection.rb
```diff
@@ -86,7 +87,26 @@ def self.wrap_with_ssl(io, tls_options = {})
...
+      if timeout
+        conn.connect_nonblock
+      else
+        conn.connect
+      end
+    rescue IO::WaitReadable
+      if IO.select([conn], nil, nil, timeout)
+        retry
+      else
+        raise Net::LDAP::LdapError, "OpenSSL connection read timeout"
```
satoryu: `Net::LDAP::LdapError` is deprecated so please use `Net::LDAP::Error` instead. 
I as a user of `net-ldap` want new exception class for timeout to let us handle cases where connection timeout happens. `lib/net/ldap/error.rb` would help you, I think.

tmm1: I switch to `SocketError`, but I can add a new `TimeoutError` or return `ETIMEDOUT` if that makes more sense.

satoryu: `Errorno::ETIMEDOUT` makes more sense in terms of compatibility since `TCPSocket` raises `Errorno::ETIMEDOUT`. 

@jch @mtodd how do you think?

tmm1: Sounds good, switched to `Errno::ETIMEDOUT`

sgringwe:
@tmm1 may i ask what prompted this PR? I recently ran into issues when using puma (rather than unicorn) for an application which uses LDAP which is resulting in timeouts in the socket#connect method when binding and searching. hoping this is related and could fix our issue!

test/test_ldap_connection.rb
```diff
@@ -291,7 +291,7 @@ def test_queued_read_setup_encryption_with_start_tls
       and_return(result2)
     mock.should_receive(:write)
     conn = Net::LDAP::Connection.new(:socket => mock)
-    flexmock(Net::LDAP::Connection).should_receive(:wrap_with_ssl).with(mock, {}).
+    flexmock(Net::LDAP::Connection).should_receive(:wrap_with_ssl).with(mock, {}, nil).
```
satoryu: @tmm1 Doesn't this mocking cover your changes on `wrap_with_ssl`, do ti? If you tested your changes work as you expected, could you share how to do it? 

I know it's really hard to reproduce the situation timeout or something network issues happens.

tmm1: I tested the change by setting a very low timeout on LDAPS connection and
make sure an exception is raised.

satoryu: ok, thanks!

tmm1:
> @tmm1 may i ask what prompted this PR? I recently ran into issues when using puma (rather than unicorn) for an application which uses LDAP which is resulting in timeouts in the socket#connect method when binding and searching. hoping this is related and could fix our issue!

Sounds like you need to set a custom `connect_timeout` on your LDAP connection, since the `DefaultConnectTimeout` is 5 seconds.

satoryu:
@tmm1 sorry for late reply. Your PR looks good. but unfortunately I cannot see Merge button :(

So anybody have a look and merge this PR ? @jch @mtodd

PR #279, opened 2016-08-23
enable TLS hostname validation

### LINKED ISSUE #258: LDAPS vulnerable to MITM - failure to validate hostname against CN or SAN in X509 Cert

tmaher:
This PR is a continuation of #259 and fixes #258. The major additions to @JPvRiel 's work are...

* Fix Vagrant integration tests - `TestBindIntegration#test_bind_timeout` relies on an iptables-blackhole'd port (tcp/8389). I switched the iptable invocation in `script/install-openldap` so that blockage extends to the hypervisor. Also I updated the docs a bit.
* New `test/fixtures/ca` files - For Vagrant testing to work, the hypervisor & guest need to both agree on a CA. The previous `cacert.pem` didn't have a corresponding private key in the repo, which made that hard. This PR adds a new CA & key, and extends `script/install-openldap` to use them.
* A bunch of TLS tests - `test/integration/test_bind.rb` already included `test_bind_tls_with_cafile`, so I added the new TLS cert verification tests there. I also refactored that file a teeny bit because of all the repetition: mostly adding some default hashes.

<strike>The first CI run is going to fail, because I'm not entirely sure what the right value is for `test_bind_tls_with_multiple_bogus_hosts`'s exception. Once the PR is created and CI runs, I should be able to get that and add it.</strike> - FIXED

/cc @jch @etdsoft


tmaher:
Ack, what happened to CI???

tmaher:
and it's back... huh

tmaher:
Note to self to check tomorrow: http://stackoverflow.com/questions/9199660/why-is-ruby-unable-to-verify-an-ssl-certificate#9238221

lib/net/ldap/connection.rb
```diff
@@ -392,12 +401,11 @@ def search(args = nil)
...
-            query_limit = (((size - n_results) < 126) ? (size -
-                                                              n_results) : 0)
-          else
-            query_limit = size
-          end
+          query_limit = if paged
+                          (((size - n_results) < 126) ? (size - n_results) : 0)
+                        else
+                          size
+                        end
```
jch: No behavioral change, just fiddling with syntax

script/install-openldap
```diff
@@ -110,6 +110,6 @@ chmod g+r /etc/ssl/private/ldap01_slapd_key.pem
 chmod o-r /etc/ssl/private/ldap01_slapd_key.pem
 
 # Drop packets on a secondary port used to specific timeout tests
-iptables -A OUTPUT -p tcp -j DROP --dport 8389
+iptables -A INPUT -p tcp -j DROP --dport 8389
```
jch: Could you provide more a link or some background for why mac os blackholes here?

script/install-openldap
```diff
@@ -48,20 +48,20 @@ chown -R openldap.openldap /var/lib/ldap
...
+export CA_KEY="/etc/ssl/private/cakey.pem"
+export CA_INFO="/etc/ssl/ca.info"
 
-sh -c "certtool --generate-privkey > /etc/ssl/private/cakey.pem"
+# If you ever need to regenerate these...
+# certtool --generate-privkey > /path/to/cakey.pem
+# certtool --generate-self-signed \
+#  --load-privkey /path/to/cakey.pem
+#  --template /path/to/ca.info
+#  --outfile /path/to/cacert.pem
```
jch: Can this be removed? Was there a previous fixture CA?

test/integration/test_bind.rb
```diff
@@ -10,7 +10,9 @@ def test_bind_timeout
     error = assert_raise Net::LDAP::Error do
       @ldap.bind(method: :simple, username: "uid=user1,ou=People,dc=rubyldap,dc=com", password: "passworD1")
     end
-    assert_equal('Connection timed out - user specified timeout', error.message)
+    msgs = ['Operation timed out - user specified timeout',
+            'Connection timed out - user specified timeout']
+    assert_send([msgs, :include?, error.message])
```
jch: This exposes the implementation too much, would it be possible to have separate tests for the two different types of errors?

lib/net/ldap/connection.rb
```diff
@@ -52,7 +52,8 @@ def open_connection(server)
     hosts.each do |host, port|
       begin
         prepare_socket(server.merge(socket: @socket_class.new(host, port, socket_opts)), timeout)
-        if encryption[:tls_options][:verify_mode] != OpenSSL::SSL::VERIFY_NONE
+        if encryption && encryption[:tls_options] &&
```
jch: Perhaps `#encryption` should always return a hash, even if no encryption is specified. This nil check feels awkward.

test/integration/test_bind.rb
```diff
@@ -64,4 +64,9 @@ def test_bind_tls_with_bad_hostname
       error.message,
     )
   end
+
+  def test_bind_tls_with_good_hostname
+    omit_if true
+    assert_true false
+  end
```
jch: What is this?

test/support/vm/openldap/README.md
```diff
@@ -27,6 +46,10 @@ $ export INTEGRATION_HOST=$ip
 
 # now run tests without having to set ENV variables
 $ time bundle exec rake
+
+# Once you're all done
+$ cd test/support/vm/openldap
+$ vagrant destroy
 ```
 
 You may need to `gem install vagrant` first in order to provision the VM.
```
jch: Earlier in this file, it lists that the vagrant gem is no longer supported:

> NOTE: The Vagrant gem (gem install vagrant) is no longer supported

Perhaps it would be good to remove this to avoid confusion?

jch: I wasn't able to run the tests according to these instructions. Here's a gist of the error https://gist.github.com/jch/48c8855edf47a58cbf953cf6adb917d0

script/install-openldap
jch: Can these be removed?

test/integration/test_bind.rb
```diff
@@ -25,18 +29,171 @@ def test_bind_anonymous_fail
...
+      method:      :start_tls,
+      tls_options: { verify_mode: OpenSSL::SSL::VERIFY_PEER,
+                     ca_file:     CA_FILE },
+    )
+    assert @ldap.bind(BIND_CREDS),
+           @ldap.get_operation_result.inspect
+  end
+
+  # The following depend on /etc/hosts hacking.
+  # We can do that on CI, but it's less than cool on people's dev boxes
```
jch: :+1:

jch:
@tmaher thanks for picking this up again! I reviewed many of the commits individually to build up an understanding because this is an area that I have not looked at for a long time. I appreciate that you also updated the documentation for running the integration tests. Unfortunately, I was not able to run it locally on my machine. Have you run into this error? https://gist.github.com/jch/48c8855edf47a58cbf953cf6adb917d0

script/install-openldap
tmaher: DigitalOcean [has a good write-up of iptables theory](https://www.digitalocean.com/community/tutorials/a-deep-dive-into-iptables-and-netfilter-architecture).

In Travis CI, the test suite executes on the integration host itself. The `-A OUTPUT` bit means that the packets are dropped before the kernel even tries sending them anywhere. In contrast, with Vagrant, the test suite runs on the hypervisor and connects to a "remote" integration host. The hypervisor is unlikely to be doing any of its own output filtering, so the packets are sent to the guest VM. The guest kernel doesn't have anything listening on tcp/8389, so it sends a RST before even consulting the `OUTPUT` chain.

With the switch to `INPUT`, on Travis, there's no effective change. TCP SYN's to 8389 will make it through the `OUTPUT` chain intact and actually be sent to localhost:8389. At that point, the `INPUT` chain for that port will drop the SYN rather than send a RST. That means from the unit tests' perspective, this is a noop.

With Vagrant, now that the guest is dropping tcp/8389 on `INPUT`, the guest kernel will stop sending RST's and just ignore the connection attempt. That brings it in line with what Travis is doing, and is what we need to do the test.

script/install-openldap
tmaher: Yes, there was a previous fixture CA, removed in this PR in caf191102f5d04cb1e4222b6b75c15a44470134e. Oddly, it was just the CA cert, no private key. Travis ignored it completely and generated its own self-signed cert/key pair on every run. Because the Vagrant tests execute on a different host, it's easier to use a fixture keypair burned into the repo. 

I left the commented-out lines in there in case the fixture CA needs to be regenerated, or in case future custom TLS tests require specific cert values. We could move the comment out into `script/generate-fixture-ca` or something if you like.

test/integration/test_bind.rb
tmaher: This is the tcp/8389 timeout check that requires the iptables `OUTPUT`/`INPUT` change. They are actually the same error. OpenSSL generates the error message text, and it provides different text on macOS:
```bash
tmaher@turing:~$ openssl s_client -connect 172.28.128.3:8389
connect: Operation timed out
connect:errno=60
```
...versus on Linux in the VM...
```
vagrant@rubyldap:~$ openssl s_client -connect 172.28.128.3:8389
connect: Connection timed out
connect:errno=110
```
This is why further down int his file I'm also rescuing two different possible exceptions for the same test: macOS vs Linux have different results. Would it make more sense for `#bind` and other bits to normalize the timeout error message, so it's the same across all platforms?

script/install-openldap
jch: > We could move the comment out into script/generate-fixture-ca or something if you like.

:+1: I think that would be more clear and prevent someone (likely me) from accidentally removing it. Maybe leave a comment pointing to that script?

test/integration/test_bind.rb
jch: > Would it make more sense for #bind and other bits to normalize the timeout error message, so it's the same across all platforms?

:+1: I like this. It'd be less surprising as a user of this library. Up to you whether you want to do it in this PR, or follow up with a separate one.

tmaher:
@jch I haven't seen that one specifically, but I have an idea. My dev box is OS X 10.11.6 (aka El Capitan), with Vagrant 1.8.5 and VirtualBox 5.1.2 (though it just told me to go update to 5.1.4).  Try this...

* In `test/support/vm/openldap` run `vagrant destroy`
* Update to latest vagrant & virtualbox. If you've got the vagrant gem, `gem uninstall vagrant` (and `rbenv rehash` if you're an rbenv'er) and `hash -r` to clear bash's cache.
* In `test/support/vm/openldap` run `vagrant plugin install vagrant-vbguest`
* `vagrant up`

The vbguest plugin should have `vagrant up` update the Guest Extensions in the 12.04 image. Since you're getting an error about the guest not being ready for communication, I'm guessing that's it. I had the plugin installed already.

tmaher:
Here's the output of a fresh `vagrant up` on my box: https://gist.github.com/tmaher/8c5bff3afd7a8427277b2585a2d7d0be

jch:
Very nice, that did it for me https://gist.github.com/jch/2f4d5de30488382fca8554956c13c41a

script/install-openldap
tmaher: Done, it's now in `script/generate-fixture-ca` and requires an explicit flag to do the destructive thing. I validated the script by using it to replace the CA/key I'd previously added to the PR. Seems fine.

test/integration/test_bind.rb
tmaher: > Up to you whether you want to do it in this PR, or follow up with a separate one.

I think next PR. I have some ideas about simplifying invocation of `Net::LDAP.new`, driven by having to wrap the whole gem more than once. Changing the exception/error message for timeouts feels like it belongs in a broader discussion about API changes for the gem.

lib/net/ldap/connection.rb
tmaher: I re-did this so it's a better copy of #259, but the nil check remains. I think you're right, it should return at least an empty hash. Next PR?

tmaher:
OK, 72ba381853e71620e9a82071eff522a144dd10df adds basic instructions to the Net::LDAP class doc on how to do cert validation, including using the OS-provided CA store.

@jch - unless there are other technical concerns (past making the API easier to use), I think this is ready. Making the API friendlier is the next PR.

chrisbloom7:
I am noticing that this was closed in August, but the last release was the month before. Any ETA on when the next release will be cut that includes this?

jch:
@chrisbloom7 thanks for the bump, released https://rubygems.org/gems/net-ldap/versions/0.16.0

krzcho:
I got a question here: I was using domain name as host and now with hostname validation it is failing due to multiple domain controllers having their own hostnames within the certificate (but within the domain itself); it seems I am no longer able to use domain as a host as I need to chose one of the controllers directly, right?

JPvRiel:
> I need to chose one of the controllers directly, right?

@krzcho that'll work, but obviously recommended against as you'll lose the redundancy of multiple DCs (much less reliable).

Sounds like someone setup a DNS based load balanced address, and if your domain admins are doing that by just adding DNS records, but not updating certs, it will fail. They're going to need to perform some custom certificate template work with the built in CA for the domain to make it work correctly/securely.

If the admins of your domain set things up correctly, in practise just pointing to `<org domain>.<tld>` should resolve to IPs of multiple DCs with LDAP and I'd assume the Microsoft CA certificate automation and templates would have subject alternative names (SAN) setup for the domain so that hostname validation can still work (google TLS/SSL load balancing and certificate SAN for more on the subject). You then don't need to point directly to a specific DC with `<dc hostname>.<org domain>.<tld>`.

To validate, compare by looking up the service records (SRV) vs just the domain. E.g.

`dig srv _ldap._tcp.<org domain>.<tld>` vs `dig <org domain>.<tld>`

However, if you're authenticating users across a forest, this tricky and you might need to only aim at global catalogs (GC) - where not all DCs are GCs (depending on how the domain was built with various roles).

krzcho:
Thank you @JPvRiel for thorough explanation; as it is corpo stuff I will probably stick to a one DC for now :(

PR #285, opened 2017-03-25
Fixed Exception: incompatible character encodings: ASCII-8BIT and UTF-8 in filter.rb

patelanuj28:
Fixed Exception: incompatible character encodings: ASCII-8BIT and UTF-8

The binary form of 5936AE79-664F-44EA-BCCB-5C39399514C6 triggers a BINARY -> UTF-8 conversion error


patelanuj28:
Can anyone look at this pull request?

vaughany:
If someone could look into this PR/fix, we sure would appreciate it.

PR #292, opened 2017-06-29
Add explicit notes around password-less simple auth

captainpete:
Hi there,

The simple auth bind example does not show how unauthenticated bind success responses can happen. Many A/D servers are configured to have successful bind when no password is supplied if the username is correct. [RFC4513 5.1.2](https://tools.ietf.org/html/rfc4513#section-5.1.2) recommends only assuming authenticated bind when password sent was present (and to improve the A/D server configuration).

At the risk of complicating the example I've created this pull request, please let me know how I can improve it!

Thanks,
Pete

### DIFF
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -814,6 +814,12 @@ class Net::LDAP
   # unrecognized usernames and incorrect passwords. Use
   # #get_operation_result to find out what happened in case of failure.
   #
+  # Note: When using the simple authentication method, some A/D configurations
+  # will return success for a bind when no password is supplied! (this is the
+  # "Unauthenticated Authentication Mechanism of Simple Bind" covered under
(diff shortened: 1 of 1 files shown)


lib/net/ldap.rb
```diff
@@ -823,7 +829,11 @@ def search(args = {})
   #  ldap.port = 389
   #  ldap.auth your_user_name, your_user_password
   #  if ldap.bind
-  #    # authentication succeeded
+  #    if your_user_password.size > 0
```
captainpete: I think this complicates the example, should `auth`, `authenticate` or `bind` check for presence of the password internally when the authentication method is `simple`?.

aurcioli-handy:
FWIW This to me is a major security oversight of the library and a big gotcha! I just stumbled upon this after realizing a blank password was accepted for an admin login on one of our internal services utilizing this library.

PR #303, opened 2018-04-16
Use Base64.strict_encode64 and SSHA256

bamthomas:
in the default ruby implementation : `Base64:encode64`, `\n` is inserted every 60 encoded characters. It is legal according to the [RFC 2045](https://www.ietf.org/rfc/rfc2045.txt) and was done for sending binary content in e-mail, where the line length is limited.

cf https://stackoverflow.com/questions/38370512/stub-random-value-in-rspec-with-secure-random

When we are using only ruby it works fine. If you are using another stack (like Spring security in java), it often uses the [RFC 4648](https://tools.ietf.org/html/rfc4648) that does not allow non ascii characters.

We cannot see it with SSHA (or others algorithms in ruby-ldap) as the salt + the hash is less than 60 chars. But if we implement longer hashes like SSHA256, there are `\n` that are inserted. 

With strict_encode64 we don't need to chomp.


bamthomas:
maybe this is like the pull #201

Issue #266, opened 2016-02-04
Undefined method `ord` for nil class


dasomx77:
Hey there, 
I am currently seeing intermittent issues with this gem; every once a while, our connection with ldap will drop and fail with this backtrace:
"/opt/shared/bundle/ruby/2.2.0/gems/net-ldap-0.11/lib/net/ldap/connection.rb:33:in `getbyte'", "/opt/shared/bundle/ruby/2.2.0/gems/net-ldap-0.11/lib/net/ber/ber_parser.rb:169:in`read_ber'", "/opt/shared/bundle/ruby/2.2.0/gems/net-ldap-0.11/lib/net/ldap/connection.rb:178:in `block in read'", "/opt/shared/bundle/ruby/2.2.0/gems/net-ldap-0.11/lib/net/ldap/instrumentation.rb:19:in`instrument'", "/opt/shared/bundle/ruby/2.2.0/gems/net-ldap-0.11/lib/net/ldap/connection.rb:177:in `read'", "/opt/shared/bundle/ruby/2.2.0/gems/net-ldap-0.11/lib/net/ldap/connection.rb:143:in`queued_read'", "/opt/shared/bundle/ruby/2.2.0/gems/net-ldap-0.11/lib/net/ldap/connection.rb:253:in `bind_simple'", "/opt/shared/bundle/ruby/2.2.0/gems/net-ldap-0.11/lib/net/ldap/connection.rb:222:in`block in bind'", "/opt/shared/bundle/ruby/2.2.0/gems/net-ldap-0.11/lib/net/ldap/instrumentation.rb:19:in `instrument'", "/opt/shared/bundle/ruby/2.2.0/gems/net-ldap-0.11/lib/net/ldap/connection.rb:219:in`bind'"

As far as I can tell, this works flawlessly most of the time since I have a couple of cron jobs that reuse the same piece of code over and over again and this only happens (very consistently) every week or may be a bit more frequently than that. The steps to reproduce this issue are literally attempting to bind. I am tempted to say perhaps that when the connection times out, either a nil object is returned to the calling function which then attempts to send it the ord command which then fails, but again, I haven't taken a look at the source code as of now.

dasomx77:
Bump? Anyone aware of what could be triggering this?

mynameisrufus:
[...]

PR #306, opened 2018-05-11
Handle nil value in GetbyteForSSLSocket::getbyte

tmccombs:
Related to  #266

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -74,7 +74,8 @@ class Net::LDAP::Connection #:nodoc:
 
   module GetbyteForSSLSocket
     def getbyte
-      getc.ord
+      c = getc
+      c && c.ord
     end
   end
 


tmccombs:
I'm not sure that this will completely fix #266. This might just result in a different error. But it seems to me that this should be the correct behavior, and it looks like at least some of the places that call this do expect to get nil back sometimes.

schaary:
thanks for the PR! Currently I am not able to push a new release, because I am on vacation an fortunately far away from my computer. I am back at the beginning of August.

dawid-sklodowski:
Any chances for merging this pull request in?

dklotz:
@schaary @mtodd Any chance a maintainer could look at this and #308 ?

tmccombs:
Any progress on this?

cjs:
This PR fixed an issue for us when interfacing with the [Duo Authentication LDAP proxy](https://duo.com/docs/authproxy-reference). It seems to be low risk and would be helpful if merged.

felixbuenemann:
I also ran into this bug when using haproxy as LDAPS proxy in front of a pool of AD LDAP servers and haproxy dropped the connection because none of the backendas where healthy.

Greg-Myers-SB:
@HarlemSquirrel Would it be possible to get a review of this fix please?
I believe a few forks exist solely to use this patch.

HarlemSquirrel (approved):
LGTM

Issue #269, opened 2016-04-04
Strange error `undefined method 'result_code'`


dblessing:
In GitLab we're seeing an occasional error when we search for a group by CN. The stack is thrown deep within net/ldap. I'm still investigating to get more information from our side, but I'm posting the stack here in hopes it'll spark some thoughts on what might be causing this. The search filter only has a group base DN and group CN.

```
2016-04-04_13:23:25.50020 2016-04-04T13:23:25.500Z 8882 TID-v9x84 WARN: NoMethodError: undefined method `result_code' for nil:NilClass 
2016-04-04_13:23:25.50022 2016-04-04T13:23:25.500Z 8882 TID-v9x84 WARN: /opt/gitlab/embedded/service/gem/ruby/2.1.0/gems/net-ldap-0.12.1/lib/net/ldap/connection.rb:450:in `block (2 levels) in search' 
2016-04-04_13:23:25.50022 /opt/gitlab/embedded/service/gem/ruby/2.1.0/gems/net-ldap-0.12.1/lib/net/ldap/connection.rb:363:in `loop' 
2016-04-04_13:23:25.50023 /opt/gitlab/embedded/service/gem/ruby/2.1.0/gems/net-ldap-0.12.1/lib/net/ldap/connection.rb:363:in `block in search' 
2016-04-04_13:23:25.50023 /opt/gitlab/embedded/service/gem/ruby/2.1.0/gems/net-ldap-0.12.1/lib/net/ldap/instrumentation.rb:19:in `instrument' 
2016-04-04_13:23:25.50023 /opt/gitlab/embedded/service/gem/ruby/2.1.0/gems/net-ldap-0.12.1/lib/net/ldap/connection.rb:352:in `search' 
2016-04-04_13:23:25.50024 /opt/gitlab/embedded/service/gem/ruby/2.1.0/gems/net-ldap-0.12.1/lib/net/ldap.rb:753:in `block (2 levels) in search' 
2016-04-04_13:23:25.50024 /opt/gitlab/embedded/service/gem/ruby/2.1.0/gems/net-ldap-0.12.1/lib/net/ldap.rb:1223:in `use_connection' 
2016-04-04_13:23:25.50025 /opt/gitlab/embedded/service/gem/ruby/2.1.0/gems/net-ldap-0.12.1/lib/net/ldap.rb:752:in `block in search' 
[...]

PR #308, opened 2018-05-15
Fix to "undefined method 'result_code'"

guilhermewop:
Related to #269

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -467,6 +467,10 @@ class Net::LDAP::Connection #:nodoc:
           end
         end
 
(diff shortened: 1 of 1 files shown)


olleolleolle:
The change in this PR raises a connection error if the interaction with the LDAP server didn't lead to a non-nil `result_pdu`.

This seems like a valid solution to the issues reported in #269.

dklotz:
@jch @schaary any chance a maintainer could take a look at this and #306 - a new release including those could probably really help those affected by #269

jch:
@dklotz sorry, I don't have the time to continue maintaining this project. I'll follow up with the other maintainers.

PR #317, opened 2019-02-14
Fix 'uninitialized constant Net::LDAP::PDU::LdapPduError' error

unagi:
This PR **fixes 'uninitialized constant Net::LDAP::PDU::LdapPduError** when unknown pdu-type occured.


test/test_ldap_connection.rb
```diff
@@ -340,6 +340,19 @@ def test_queued_read_bind_sasl
     assert result.success?
     assert_equal 2, result.message_id
   end
+
+  def test_invalid_pdu_type
+    options = {
+      app_tag: Net::LDAP::PDU::SearchResult,
```
satoryu: Is this required? not used in this test case.

lib/net/ldap/pdu.rb
```diff
@@ -123,7 +123,7 @@ def initialize(ber_object)
     when ExtendedResponse
       parse_extended_response(ber_object[1])
     else
-      raise LdapPduError.new("unknown pdu-type: #{@app_tag}")
+      raise Net::LDAP::PDU::Error, "unknown pdu-type: #{@app_tag}"
```
satoryu: Thanks!

test/test_ldap_connection.rb
unagi: That's right!
I remove this parameter.

unagi:
@mtodd @satoryu  ping

PR #325, opened 2019-07-11
Include 'net/ldap/dn' class

cmdrclueless:
The class `Net::LDAP::DN` is not being required by default in the `net/ldap` module.


mtodd:
Thank you for opening this PR!

`Net::LDAP::DN` is not used by the core `Net::LDAP` facility; it's intended to be an opt in, standalone utility class.

Given it's not a requirement for the core operation, I'm going to close this PR as it's not needed and should not be required by default.

PR #326, opened 2019-08-06
Make the `generate()` method more idiomatic...

pvdb:
... and fix the following Ruby warning in the process:

```
lib/net/ldap/password.rb:22: warning: assigned but unused variable - attribute_value
```

Doesn't change anything functionally, and the tests still pass!  :smiley:


lib/net/ldap/password.rb
```diff
@@ -19,20 +19,18 @@ class << self
...
       when :md5
-         attribute_value = '{MD5}' + Base64.encode64(Digest::MD5.digest(str)).chomp!
+         '{MD5}' + Base64.encode64(Digest::MD5.digest(str)).chomp!
       when :sha
-         attribute_value = '{SHA}' + Base64.encode64(Digest::SHA1.digest(str)).chomp!
+         '{SHA}' + Base64.encode64(Digest::SHA1.digest(str)).chomp!
       when :ssha
          salt = SecureRandom.random_bytes(16)
-         attribute_value = '{SSHA}' + Base64.encode64(Digest::SHA1.digest(str + salt) + salt).chomp!
+         '{SSHA}' + Base64.encode64(Digest::SHA1.digest(str + salt) + salt).chomp!
```
HarlemSquirrel, line 30: I think https://ruby-doc.org/stdlib/libdoc/base64/rdoc/Base64.html#method-i-strict_encode64 would make this even more simple and not require calling `#chomp`

pvdb: Thanks for the feedback!

The changes in this PR _(as well as the similar ones in #327 and #328)_ focussed on suppressing the Ruby warnings, whilst shying away from making any functional changes.

I'll gladly to the functional refactoring you suggested as well, but how do you feel about merging this PR as is, and I'll follow up immediately afterwards with the further more functional refactoring?

HarlemSquirrel: OK, that works for me.

pvdb: Done in #362!

PR #328, opened 2019-08-06
Make the `instrument()` method more idiomatic...

pvdb:
... and fix the following Ruby warning in the process:

```
lib/net/ldap/instrumentation.rb:15: warning: shadowing outer local variable - payload
```

Doesn't change anything functionally, and the tests still pass!  :smiley:


pvdb:
Thanks for merging!

PR #331, opened 2019-11-12
Net::LDAP#open does not cache bind result

vroldanbet:
We identified that clients cannot safely rely on `Net::LDAP#get_operation_result` when using `Net::LDAP#open` because `@result` is not set.

As a consequence, clients calling `Net::LDAP#get_operation_result` would get the previous cached `@result`. Yikes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ publish/
 Gemfile.lock
 .bundle
 bin/
+.idea
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -712,7 +712,7 @@ class Net::LDAP
       begin
         @open_connection = new_connection
         payload[:connection] = @open_connection
-        payload[:bind]       = @open_connection.bind(@auth)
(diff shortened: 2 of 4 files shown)


vroldanbet:
In https://github.com/ruby-ldap/ruby-net-ldap/pull/331/commits/8135bbe7ef5fe8beb5767d69c9467cb7b401cde7 we can see the result of the test, which fails to pass.

In https://github.com/ruby-ldap/ruby-net-ldap/pull/331/commits/dd2a880bcfc2c4433f8de855eb24a455039f396b the test goes green after introducing the fix in https://github.com/ruby-ldap/ruby-net-ldap/pull/331/commits/cb4b48cb5012ed25176f978900b4357a858819e6

vroldanbet:
I'm addressing CI build in https://github.com/ruby-ldap/ruby-net-ldap/pull/333

vroldanbet:
Superseded by https://github.com/ruby-ldap/ruby-net-ldap/pull/334, which contains the fixed CI build

Issue #228, opened 2015-10-19
Implement GSS_SPNEGO Adapter as a gem


satoryu:
In discussion of #226, GSS_SPENEGO authentication adapter must be implemented as a gem to care users who use this authentication method. 

CC: @jch

satoryu:
@jch @mtodd Could you please assign me to this issue?

satoryu:
Referring [the guide](http://guides.rubygems.org/name-your-gem/), the gem name will be `net-ldap-auth_adapter-gss_spnego`.  how about this?

jch:
That gem name is a mouthful, but I think it's clear what it's supposed to me. We can also link to it from the README so it's easier for others to find.

satoryu:
I could not find the repository of the gem 'ntlm-http' on which `gss_spnego` depends... ;(

jch:
Did it get yanked? If the implementation isn't too long, you could find the code and just inline it into your gem.

tmaher:
@satoryu - it looks like it's on rubygems now: https://rubygems.org/gems/ntlm-http/versions/0.1.1

repo appears to be https://github.com/pyu10055/ntlm-http though it's a couple minor revs ahead of what's on RubyGems (0.1.3.3 vs 0.1.1)

larskanis:
This issue is solved by the gem `net-ldap-gss-spnego`, so that this issue can be closed and the provisional file for gss_spnego can be removed from net-ldap gem. See #343

PR #343, opened 2020-01-31
Remove provisonal gss-spnego and add documentation for net-ldap-gss-pnego gem

fwininger:
Remove provisonal gss-spnego and add documentation for net-ldap-gss-pnego gem


larskanis:
I support this PR, hence removing the provisional code. A dedicated gem with it's own dependencies was implemented and released [here](https://github.com/fwininger/ruby-net-ldap-gss-spnego), so that #228 is fixed that way. And it's troublesome to have two gems that provide one and the same file, so that the require is ambitious:
```
require 'net/ldap/auth_adapter/gss_spnego'
```

PR #349, opened 2020-02-14
Don't verify hostname when verify_hostname is false in tls_options

jurta:
https://ruby.github.io/openssl/OpenSSL/SSL/SSLContext.html documents the option `verify_hostname` as whether to check the server certificate is valid for the hostname.

But when I set `verify_hostname` to `false` in `tls_options`, it has no effect, it still reports the error "hostname does not match the server certificate".

Then I found this is the result of #259 but the assumption is wrong that users who don't care about hostname validation should set `verify_mode` to `OpenSSL::SSL::VERIFY_NONE`. This disables the certificate validation completely. Maybe the reason why #259 didn't add a check for `verify_hostname` is because `verify_hostname` was added later in https://github.com/ruby/openssl/pull/60 (i.e. the same year but a few months later).

So for more fine-grained configuration to disable only hostname verification without disabling certificate validation, here is this pull request.

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -53,8 +53,10 @@ class Net::LDAP::Connection #:nodoc:
         prepare_socket(server.merge(socket: @socket_class.new(host, port, socket_opts)), timeout)
         if encryption
           if encryption[:tls_options] &&
-             encryption[:tls_options][:verify_mode] &&
-             encryption[:tls_options][:verify_mode] == OpenSSL::SSL::VERIFY_NONE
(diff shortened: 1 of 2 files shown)


secos:
Any updates on this?

zeroSteiner:
I merged the changes from `master` into this branch so the tests can be run and it looks like there's some failures. I think one is due to `127.0.0.1` not being the IP address of the LDAP server in the CI environment. I'd recommend changing it to `@ldap.host = Resolv.getaddress(INTEGRATION_HOSTNAME)`. I ran another test with that in place and got a new error though which is when I stopped investigating.

PR #342, opened 2019-12-27
add explicit ** to silence Ruby 2.7 warning

jethrodaniel:
Closes https://github.com/ruby-ldap/ruby-net-ldap/issues/341

PR #350, opened 2020-03-24
lib/net/ldap/connections.rb: use ** to pass a keyword parameter

lucaskanashiro:
In Socket.tcp call is passed a keyword parameter as is and while I've
been testing a library that relies on it I got the following warning:

net/ldap/connection.rb:709: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call

In ruby 2.7 is preferred to use ** to pass keyword arguments in method's
call.


HarlemSquirrel:
Thanks but we've got this taken care of now with  #342

PR #352, opened 2020-03-30
Fix unknow pdu Error and show the Ber class

ggalancs:


lib/net/ldap/pdu.rb
```diff
@@ -84,7 +84,8 @@ class Error < RuntimeError; end
 
   def initialize(ber_object)
     begin
-      @message_id = ber_object[0].to_i
+      id =  ber_object[0]
```
jethrodaniel: what's the point of this 2 line change?

ggalancs: No one, sorry, I try to understand the ber object and I forgot that lines

lib/net/ldap/pdu.rb
```diff
@@ -84,7 +84,7 @@ class Error < RuntimeError; end
 
   def initialize(ber_object)
     begin
-      @message_id = ber_object[0].to_i
+      @message_id = ber_object[0]
```
HarlemSquirrel, line 87: What's the benefit of removing this coercion?

HarlemSquirrel: @ggalancs :point_up:

lib/net/ldap/version.rb
```diff
@@ -1,5 +1,5 @@
 module Net
   class LDAP
-    VERSION = "0.16.2"
+    VERSION = "0.16.2.3"
```
HarlemSquirrel, line 3: Why change the version here?

lib/net/ldap/pdu.rb
gwillcox-r7: @HarlemSquirrel Believe this was to prevent a few casting issues when the item could not be converted into an integer. Run into a few of these myself when testing.

PR #356, opened 2020-05-14
Ability to add socket_class while instantiating net_ldap connection

dhivyada:


dhivyada:
Removed the accidentally committed gem. Thank you for the feedback

dhivyada:
@jethrodaniel : Force pushed the socket changes alone to keep the commit history clean

michaelfranzl:
This is a very useful feature IMO. This would allow a socket to be configured beyond the connection timeout, e.g. by binding it to a specific local interface and port. Are there plans to merge this?

zeroSteiner:
I merged this into a testing branch to run the CI test suite and got a bunch of failures. I think the best path forward on this would be to rebase the changes onto master and then get the tests passing. Once that's done I can test this with a custom socket class and get this landed. Thanks!

Issue #313, opened 2018-08-07
rfc2696 paging not working when server returns additional, empty pagedResultsControl items


martindorey:
I've got a customer whose LDAP server repeatably returns a result like this:

    LDAPMessage searchResDone(2) success [126 results]
        messageID: 2
        protocolOp: searchResDone (5)
            searchResDone
                resultCode: success (0)
                matchedDN:
                errorMessage:
        [Response To: 28]
        [Time: 1.821359773 seconds]
        controls: 4 items
            Control
                controlType: 1.2.840.113556.1.4.319 (pagedResultsControl)
                SearchControlValue
                    size: 0
                    cookie: 006f753d7573657273006f753d70656f706c65006f753d65...
            Control
                controlType: 1.2.840.113556.1.4.319 (pagedResultsControl)
                SearchControlValue
                    size: 0
                    cookie: <MISSING>
            Control
                controlType: 1.2.840.113556.1.4.319 (pagedResultsControl)
                SearchControlValue
                    size: 0
                    cookie: <MISSING>
            Control
                controlType: 1.2.840.113556.1.4.319 (pagedResultsControl)
                SearchControlValue
                    size: 0
                    cookie: <MISSING>

That is to say one containing four pagedResultsControls, only the first of which has a non-empty cookie.  I don't think it's important but, for the record, the query looked like this:

    LDAPMessage searchRequest(2) "DC=example,DC=com" wholeSubtree
        messageID: 2
        protocolOp: searchRequest (3)
            searchRequest
                baseObject: DC=example,DC=com
                scope: wholeSubtree (2)
                derefAliases: neverDerefAliases (0)
                sizeLimit: 0
                timeLimit: 0
                typesOnly: False
                Filter: (objectClass=posixAccount)
                    filter: equalityMatch (3)
                        equalityMatch
                            attributeDesc: objectClass
                            assertionValue: posixAccount
                attributes: 7 items
                    AttributeDescription: uid
                    AttributeDescription: userPassword
                    AttributeDescription: uidNumber
                    AttributeDescription: gidNumber
                    AttributeDescription: cn
                    AttributeDescription: homeDirectory
                    AttributeDescription: loginShell
        controls: 1 item
            Control
                controlType: 1.2.840.113556.1.4.319 (pagedResultsControl)
                criticality: False
                SearchControlValue
                    size: 126
                    cookie: <MISSING>

[...]

Issue #36, opened 2012-04-10
Wrong paged result from search


cfbrobak:
Environment: Windows XP, Java 1.6.0_29, JRuby 1.6.7 and net-ldap 0.3.1. JRuby is running in Ruby 1.9 mode.

Scenario: Searching for computers in our Windows Active Directory.

As I understand it, if the search result contains more than 126 (hardcoded page size in net/ldap.rb) entries, which it always does in my case, the server returns a cookie indicating a paged result (RFC 2696). The cookie should be sent back by the client when requesting more entries from the result. This does not seem to work. The result is always 126 bigger than it should be and those 126 seem to be duplicates of some already in the result.

Another problem: If the result contains more than 1000 entries a "Size Limit Exceeded" error is returned.

My guess is that net-ldap somehow changes the cookie before sending it back to the server.

If I switch to net-ldap 0.2.2, it seems to work as it should.

saimonmoore:
this is still an issue for me. Even though the page size is set to 126 and the possible result set is generally much larger than 1000 entries (I set the size to a very large number as I want the whole result set) the behaviour I'm seeing with AD is:

- a first page containing 126 results
- a 2nd page up until 1126 results
- then Size Limit exceeded error

as if it was not actually doing any paging (126 results at a time).

In ruby there doesn't seem to be any other project as active as this one but this is a real blocker for us.

@jch can you please comment on this? I'm using master (8a18267) branch and seeing this behaviour...

saimonmoore:
And I have just confirmed that 0.2.2 works fine (at least with paging)

jch:
@saimonmoore thanks for bumping this thread and the helpful reproduction list. I wasn't a maintainer back around 0.2.2, so I'll need to take some time to dig into the differences. There's been a lot of changes since then, so it may be worth revisiting this from scratch as well.

It sounds like you have a good environment for reproducing this. Would you be interested in opening a PR to attempt to fix this? I'm not very familiar with AD and don't have a local setup, so it would speed up development if you got something started. I'm happy to code review and follow up.

cc @meastman also in case you're looking for something to hack on during downtime

saimonmoore:
@jch  basically it all started to break with the introduction of utf8 encoding...

I will try and see if I have more time to track down the exact change...

jch:
Arrrgh, UTF8! :+1: to digging in.

arindam2:
[...]

PR #369, opened 2020-07-21
Fix paged search rfc2696_cookie for Microsoft Active Directory

amuta:
When building the `cookie` of a paged search we are currently calling `to_ber` when we should be calling `to_ber_bin` to avoid encoding it to UTF-8 and 'breaking' the cookie format. When looking at the method `to_ber_bin`, it seems to be build with this in mind.
```
> lib/net/ldap/connection.rb:20
# Converts a string to a BER string but does *not* encode to UTF-8 first.
# This is required for proper representation of binary data for Microsoft
# Active Directory
def to_ber_bin(code = 0x04)
  [code].pack('C') + length.to_ber_length_encoding + self
end
```

Without this change, when trying to do a paged search the behavior is the same as described in this issue: [#36 Wrong paged result from search](https://github.com/ruby-ldap/ruby-net-ldap/issues/36)

https://github.com/ruby-ldap/ruby-net-ldap/issues/36#issuecomment-156068141
> Even though the page size is set to 126 and the possible result set is generally much larger than 1000 entries (I set the size to a very large number as I want the whole result set) the behaviour I'm seeing with AD is:
> 
> a first page containing 126 results
> a 2nd page up until 1126 results
> then Size Limit exceeded error
> as if it was not actually doing any paging (126 results at a time).

The issue is very likely to be related [#313 rfc2696 paging not working when server returns additional, empty pagedResultsControl items](https://github.com/ruby-ldap/ruby-net-ldap/issues/313), because after sending the broken cookie the AD still reponds with what would be the first page after the initial query but wont send the cookie required for continuing the paged search.

The comment below seems to imply that this issue was known and disabling paging was a solution, but the paging is not disabled and it breaks silently.
```
> lib/net/ldap/connection.rb:241
 # rfc2696_cookie sometimes contains binary data from Microsoft Active Directory
 # this breaks when calling to_ber. (Can't force binary data to UTF-8)
 # we have to disable paging (even though server supports it) to get around this...
```

I've only tested this against Microsoft Active Directory, and it enabled me to used the paged search.


HarlemSquirrel:
@amuta could you please update this from the latest master branch when you have a moment?

amuta:
@HarlemSquirrel sorry for taking soooo long, but I rebased it and changed it with the code that we are working with. We are using this change at our org for more than 1 year and it is working, but if anyone can try it out with Active Directory and a search with >1.2k results it should be evident.

amuta:
The tests are breaking because of lint rules, is it possible to make the linter skip that method?

HarlemSquirrel:
@amuta This should be refactored but for now we could up the limit to 117 in 
https://github.com/ruby-ldap/ruby-net-ldap/blob/626c312b49339eb71b1013ca35e406905e6ca1b1/.rubocop_todo.yml#L207-L209

I don't have access to any Active Directory to test this myself and don't work with LDAP at all right now so we really should get someone to try this out on a real directory before merging in as well.

Are you able to write a test for this use case?

PR #370, opened 2020-07-22
WIP Add connection pool option

HarlemSquirrel:
```rb
# Example usage
ldap_with_pool = Net::LDAP.new(
  host: host_name,
  port: 389, 
  connection_pool: { 
    size: 4, # Maximum connections to open
    timeout: 15 # Number of seconds to wait for a connection to free up before raising a timeout error
  }
)
```

See https://github.com/mperham/connection_pool#usage


net-ldap.gemspec
```diff
@@ -29,6 +29,8 @@ the most recent LDAP RFCs (4510-4519, plutions of 4520-4532).}
   s.required_ruby_version = ">= 2.0.0"
   s.summary = %q{Net::LDAP for Ruby (also called net-ldap) implements client access for the Lightweight Directory Access Protocol (LDAP), an IETF standard protocol for accessing distributed directory services}
 
+  s.add_runtime_dependency("connection_pool", "~> 2.2")
```
HarlemSquirrel, line 32: I'm not sure if we want to make this a requirement or allow users to opt in by adding the gem in their Gemfile. If we do that, then we have to handle when the library is missing.

lib/net/ldap.rb
```diff
@@ -553,6 +555,10 @@ def initialize(args = {})
     @force_no_page = args[:force_no_page] || DefaultForceNoPage
     @encryption = normalize_encryption(args[:encryption]) # may be nil
     @connect_timeout = args[:connect_timeout]
+    if args[:connection_pool].is_a? Hash
+      options = { size: 5, timeout: (@connection_timeout || 5) }.merge args[:connection_pool]
```
HarlemSquirrel, line 559: I'm not sure if we should default the timeout here in this way. The time to wait for a connection to free up is not quite the same as waiting for a single connection to be made. Perhaps we should default so some higher value here like 30 seconds.

Issue #378, opened 2020-09-13
ldapserver.rb:  Unsupported object type: id=96 (Net::BER::BerError)


dup2:
I'm trying to use your testserver as the base for project to support some simple ldap queries.

However, with every client (ldapsearch and this ruby net-ldap library) I always get the following exception on the server which then terminates:
net-ldap-0.16.2/lib/net/ber/ber_parser.rb:93:in `parse_ber_object': Unsupported object type: id=96 (Net::BER::BerError)

Any ideas about this?

tatsuya-ogawa:
LdapServerAsnSyntax need compiled.
So please try to replace LdapServerAsnSyntax like this
```
LdapServerAsnSyntax = Net::BER.compile_syntax(
      {
          :application => {
              :constructed => {
                  0 => :array, # LDAP BindRequest
                  3 => :array # LDAP SearchRequest
              },
              :primitive => {
                  2 => :string, # ldapsearch sends this to unbind
              },
          },
          :context_specific => {
              :primitive => {
                  0 => :string, # simple auth (password)
                  7 => :string # present filter
              },
              :constructed => {
                  3 => :array # equality filter
              },
          },
      }
  )
```

dup2:
Thanks, that worked!

PR #379, opened 2020-09-15
fix LdapServerAsnSyntax compile

tatsuya-ogawa:
#378


HarlemSquirrel (changes requested):
Thank you for submitting this. So you have a moment to fix the lints rubocop found?

HarlemSquirrel (commented):
Also, why move this down the page?

HarlemSquirrel (approved):
Thank you!

PR #380, opened 2020-09-23
remove a circular require

jethrodaniel:
If applied, this commit removes some circular `require`s.

Prior to this change, we had the following circular `require`s

```
require_relative 'entry' unless defined? Net::LDAP::Entry     # dataset
require_relative 'dataset' unless defined? Net::LDAP::Dataset # entry
```

This works (both classes need each other in methods), but it's unnecessary, since calling `require` twice does nothing, since
`$LOADED_FEATURES` has already been updated.

This change moves those `require`s to the toplevel, and removes the `defined?` check.

---

No tests were added, since the existing tests passed without issue.

### DIFF
--- a/lib/net/ldap/dataset.rb
+++ b/lib/net/ldap/dataset.rb
@@ -1,3 +1,5 @@
+require_relative 'entry'
+
 # -*- ruby encoding: utf-8 -*-
 ##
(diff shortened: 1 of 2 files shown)


HarlemSquirrel (approved):
Thanks you!

PR #381, opened 2020-10-14
Implement '==' operator for entries

### LINKED ISSUE #372: Implement ==
I think we should be able to compare entries with `==` just [like we do with hashes](https://ruby-doc.org/core/Hash.html#method-i-3D-3D).

```rb
Hash.new(dn: 'dc=example,dc=com') == Hash.new(dn: 'dc=example,dc=com')
# => true

Net::LDAP::Entry.new(dn: 'dc=example,dc=com') == Net::LDAP::Entry.new(dn: 'dc=example,dc=com')
# => false
```

erlliam:
Hey @HarlemSquirrel I was going through your profile and saw this repository. This issue seemed like something I could solve.

Resolves #372

### DIFF
--- a/lib/net/ldap/entry.rb
+++ b/lib/net/ldap/entry.rb
@@ -196,4 +196,8 @@ class Net::LDAP::Entry
     sym.to_s[-1] == ?=
   end
   private :setter?
+
+  def ==(other)
+    return other.instance_of?(self.class) && @myhash == other.to_h
+  end
 end # class Entry
--- a/lib/net/ldap/entry.rb
+++ b/lib/net/ldap/entry.rb
@@ -198,6 +198,6 @@ class Net::LDAP::Entry
   private :setter?
 
   def ==(other)
-    return other.instance_of?(self.class) && @myhash == other.to_h
+    other.instance_of?(self.class) && @myhash == other.to_h
   end
 end # class Entry
--- a/test/test_entry.rb
+++ b/test/test_entry.rb
@@ -54,6 +54,17 @@ class TestEntry < Test::Unit::TestCase
     duplicate.delete(:sn)
     assert_not_equal duplicate, @entry.to_h
   end
+
+  def test_equal_operator
+    entry_two = Net::LDAP::Entry.new 'cn=Barbara,o=corp'
+    assert_equal @entry, entry_two
+
+    @entry['sn'] = 'Jensen'
+    assert_not_equal @entry, entry_two
+
+    entry_two['sn'] = 'Jensen'
+    assert_equal @entry, entry_two
+  end
 end
 
 class TestEntryLDIF < Test::Unit::TestCase


lib/net/ldap/entry.rb
```diff
@@ -196,4 +196,8 @@ def setter?(sym)
     sym.to_s[-1] == ?=
   end
   private :setter?
+
+  def ==(other)
+    return self['dn'] == other['dn']
```
HarlemSquirrel, line 201: This is a great start but I think we also need to ensure that all the attributes are the same.

HarlemSquirrel (changes requested):

lib/net/ldap/entry.rb
```diff
@@ -196,4 +196,8 @@ def setter?(sym)
     sym.to_s[-1] == ?=
   end
   private :setter?
+
+  def ==(other)
+    return @myhash == other.to_h
```
esparta, line 201:
```suggestion
    @myhash == other.to_h
```

Also, should this method check if the `other` is the same class?

erlliam:
Should the equal operator work for sub classes of entries ? I was going to write `other.class == self.class` instead of `instance_of?`, but I felt like the latter is probably the better choice. Is there any difference between those two?

lib/net/ldap/entry.rb
HarlemSquirrel: While we don't really have an exposed way to create objects that inherit from Net::LDAP::Entry I would say those would be OK be OK to consider equal as far as Class goes. For [Ruby Hash objects](https://ruby-doc.org/core-2.7.2/Hash.html#method-i-3D-3D) `==` Returns true if all of the following are true:

> - object is a Hash object.
> - hash and object have the same keys (regardless of order).
> - For each key key, hash[key] == object[key].

HarlemSquirrel:
While we don't really have an exposed way to create objects that inherit from Net::LDAP::Entry I would say those would be OK be OK to consider equal as far as Class goes. For [Ruby Hash objects](https://ruby-doc.org/core-2.7.2/Hash.html#method-i-3D-3D) `==` Returns true if all of the following are true:

> - object is a Hash object.
> - hash and object have the same keys (regardless of order).
> - For each key key, hash[key] == object[key].

lib/net/ldap/entry.rb
```diff
@@ -196,4 +196,8 @@ def setter?(sym)
     sym.to_s[-1] == ?=
   end
   private :setter?
+
+  def ==(other)
+    return other.instance_of?(self.class) && @myhash == other.to_h
```
HarlemSquirrel, line 201: We don't need an explicit return here.
```suggestion
    other.instance_of?(self.class) && @myhash == other.to_h
```

HarlemSquirrel:
@erlliam `other.class == self.class` and `other.instance_of?(self.class)` are doing essentially the same thing. The late does feel more to the spirit of Ruby where we often prefer what sounds or looks more "natural".

HarlemSquirrel (approved):
Nice work!

Issue #382, opened 2020-11-20
undefined method `write' for nil:NilClass


railsstarter:
net-ldap version: 0.16.0
Problem:  _While doing socket.write at https://github.com/ruby-ldap/ruby-net-ldap/blob/release-0.16.0/lib/net/ldap/connection.rb#L261 fails with exception: "undefined method `write' for nil:NilClass"_

Looking at the error, it seems that socket from https://github.com/ruby-ldap/ruby-net-ldap/blob/release-0.16.0/lib/net/ldap/connection.rb#L688 is coming as nil.

HarlemSquirrel:
Fixed by #383

PR #383, opened 2020-11-23
fix for undefined method for write exception

railsstarter:
Opening the PR for issue: https://github.com/ruby-ldap/ruby-net-ldap/issues/382

- Method#socket returns @conn sometimes as nil because it is only checking that it is defined or not. (a defined @conn can be nil when it is dead in between)
- checking that along with its definition it should not be nil too.

 Date:      Mon Nov 23 16:27:38 2020 +0530
 Committer: Kanika Gupta <railsstarter@github.com>


sukeerthiadiga:
+1

HarlemSquirrel (approved):
Nice find, thank you!

PR #385, opened 2021-01-13
Fixed shebang of bash

mfechner:
Not every OS has bash in the path used before. Let `/usr/bin/env` search for the correct path of bash.


HarlemSquirrel (approved):
OK thanks

PR #392, opened 2021-04-16
Fix circular require while loading lib/net/ldap/entry.rb and lib/net/ldap/dataset.rb

krists:
This change removes circular relative_requires from lib/net/ldap/entry.rb and lib/net/ldap/dataset.rb.
Require warnings can be seet at https://github.com/ruby-ldap/ruby-net-ldap/runs/1831714744
These requires  are not necessary as these files are already required in lib/net/ldap.rb file.


HarlemSquirrel (approved):
Nice fix! Thank you!

Issue #405, opened 2022-06-07
Add SNI support for Google Secure LDAP


jpdasma:
Using `net/ldap` with `ldap.google.com:636`:

```
require 'net/ldap'

ldap = Net::LDAP.new  :host => 'ldap.google.com',
                      :port => 636,
                      :encryption => :simple_tls,
                      :auth => {
                        :method => :simple,
                        :username => 'test',
                        :password => 'test',
                      }
```

 will result in this error:

```
`open_connection': hostname "ldap.google.com" does not match the server certificate (Net::LDAP::Error)
```

This is because of Google Secure LDAP's behavior: https://support.google.com/a/answer/9190869?hl=en

> The Secure LDAP service requires a TLS client that supports and initiates a TLS session using SNI (Server Name Indication). If the TLS client does not support SNI, then the TLS server (ldap.google.com) returns a self-signed certificate that will not pass CA validation checks, to indicate that SNI is required.

[To use SNI](http://ruby.github.io/openssl/OpenSSL/SSL/SSLSocket.html#method-i-hostname-3D) we have to add a call to `conn.hostname` in this part of the code (before calling `conn.connect`): https://github.com/ruby-ldap/ruby-net-ldap/blob/master/lib/net/ldap/connection.rb#L98

I tested this by hard coding `conn.hostname` to `ldap.google.com`:

```
@@ -95,6 +95,7 @@
     ctx.set_params(tls_options) unless tls_options.empty?

     conn = OpenSSL::SSL::SSLSocket.new(io, ctx)
+    conn.hostname = "ldap.google.com"

     begin
       if timeout
```

This fixed the error with regards to the certificate.

I'm planning on submitting a PR for this, but I figured I might open an issue first to see if I didn't miss anything.

PR #406, opened 2022-06-07
Add support to use SNI

jpdasma:
Related to https://github.com/ruby-ldap/ruby-net-ldap/issues/405

### DIFF
--- a/lib/net/ldap/connection.rb
+++ b/lib/net/ldap/connection.rb
@@ -33,9 +33,10 @@ class Net::LDAP::Connection #:nodoc:
   def prepare_socket(server, timeout=nil)
     socket = server[:socket]
     encryption = server[:encryption]
+    hostname = server[:host]
 
     @conn = socket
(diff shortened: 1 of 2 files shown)


jpdasma:
I'm thinking if it's better to have an option to use SNI or just always enable it?

The default `openssl` binary uses SNI by default.

jpdasma:
I'll fix the broken tests first.

jpdasma:
The failing test should be fixed now. I also tested with `db.debian.org:636`.

From some of my readings, SNI is an optional TLS extension, so it should be safe to be enabled by default.

Although we might still want to test it to other LDAP server with TLS.

jpdasma:
@HarlemSquirrel based on my tests, I think this should be safe to merge. I didn't encountered any issues with OpenLDAP and Windows LDAP.

HarlemSquirrel (approved):

HarlemSquirrel:
Thank you!

Issue #398, opened 2021-10-12
Net::LDAP::DN.escape() throws when attempting to escape attr value beginning with `#`


bgraves-lo:
`Net::LDAP::DN.escape()` is meant to adhere to https://datatracker.ietf.org/doc/html/rfc2253#section-2.4, which defines the convention for escaping attribute values.

Here's the related code:
https://github.com/ruby-ldap/ruby-net-ldap/blob/d6bb5c8d694662e36317f21d96cd6b02db263b6d/lib/net/ldap/dn.rb#L192-L216

The code properly escapes the special characters included in the `ESCAPES` hash, handling this case from the RFC:
```
    o   one of the characters ",", "+", """, "\", "<", ">" or ";"
```

But the problem occurs with the special cases involving '#' and space:
```
    o   a space or "#" character occurring at the beginning of the
        string

    o   a space character occurring at the end of the string
```

Space and '#' aren't included in that hash, so if `ESCAPE_RE` matches '^#', for instance, the lookup of `ESCAPES['#']` returns nil, which causes `"\\" + ESCAPES[char]` to throw a `TypeError (no implicit conversion of nil into String)`.

A potential workaround:
```
  def self.escape(string)
    string.gsub(ESCAPE_RE) { |char| "\\" + (ESCAPES[char] || char) }
  end
```

HarlemSquirrel:
Thank you for opening this issue! You have described this very well and I would be happy to review a pull request for this fix!

bgraves-lo:
This fell off my radar till now 😅 , just submitted a PR.

PR #408, opened 2022-09-16
Fix escaping of # and space in attrs

bgraves-lo:
Addresses #398: `Net::LDAP::DN.escape()` raises a `TypeError (no implicit conversion of nil into String)` when attempting to escape a space or # at the beginning of the attribute, or a space at the end.


HarlemSquirrel (approved):
Thank you for taking the time to open this big fix! Nice refctors!

PR #415, opened 2023-05-25
Document `connect_timeout` in Constructor Details

alexjfisher:
Previously, this was only documented in the `Overview` section and missing from https://www.rubydoc.info/github/ruby-ldap/ruby-net-ldap/Net%2FLDAP:initialize


HarlemSquirrel (approved):
Thank you for your contribution!

PR #422, opened 2024-01-08
Update TruffleRuby version on CI to 23.1.1 (latest)

andrykonchin:
CI fails right now on TruffleRuby 22 on `bundle check` command ([example](https://github.com/ruby-ldap/ruby-net-ldap/actions/runs/7400222333/job/20133406447)). The issue is fixed in TruffleRuby 23.


HarlemSquirrel:
Thanks for this but I've fixed this by using `truffleruby:stable` image now.

PR #396, opened 2021-06-11
Add support for ldapwhoami (RFC4532)

a7b81a9086:
Hi,
this is my first attempt at contributing to an open source project, so if anything is missing feel free to tell me. I am trying to close issue #395 with this pr.

I ran `rake rubotest` and it just complained about the `Net::LDAP` being too long, I think the code should be easy enough, as it is just copied from other functions for the most part.  Other than that I added one number to the BER-parser that was missing.

PR #425, opened 2024-02-23
Add support for ldapwhoami (RFC4532) (now with tests)

zeroSteiner:
I'm hoping to see #396 completed by adding the tests that were requested. In doing so I also fixed a problem I noticed where it was not always returning the response string.

Tested with Active Directory and OpenLDAP.

### DIFF
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -311,7 +311,7 @@ class Net::LDAP
     0 => :array, # RFC-2251 Control and Filter-AND
     1 => :array, # SearchFilter-OR
     2 => :array, # SearchFilter-NOT
-    3 => :array, # Seach referral
+    3 => :array, # Search referral
     4 => :array, # unknown use in Microsoft Outlook
     5 => :array, # SearchFilter-GE
     6 => :array, # SearchFilter-LE
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -306,7 +306,6 @@ class Net::LDAP
     3 => :string, # SearchFilter-extensible
     4 => :string, # SearchFilter-extensible
     7 => :string, # serverSaslCreds
-    11 => :string, # responseValue
   }
   constructed = {
     0 => :array, # RFC-2251 Control and Filter-AND
--- a/lib/net/ldap.rb
+++ b/lib/net/ldap.rb
@@ -306,6 +306,7 @@ class Net::LDAP
     3 => :string, # SearchFilter-extensible
     4 => :string, # SearchFilter-extensible
     7 => :string, # serverSaslCreds
+    11 => :string, # responseValue
   }
   constructed = {
     0 => :array, # RFC-2251 Control and Filter-AND
@@ -341,6 +342,7 @@ class Net::LDAP
 
   StartTlsOid = '1.3.6.1.4.1.1466.20037'
   PasswdModifyOid = '1.3.6.1.4.1.4203.1.11.1'
+  WhoamiOid = '1.3.6.1.4.1.4203.1.11.3'
 
   # https://tools.ietf.org/html/rfc4511#section-4.1.9
   # https://tools.ietf.org/html/rfc4511#appendix-A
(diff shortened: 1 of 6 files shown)


zeroSteiner:
I was just wondering if there was anything I could do to help this along?

lib/net/ldap.rb
```diff
@@ -325,7 +325,7 @@ class Net::LDAP
 
   universal = {
     constructed: {
-      107 => :array, #ExtendedResponse (PasswdModifyResponseValue)
+      107 => :string, # ExtendedResponse
```
HarlemSquirrel, line 328: Why change this?

lib/net/ldap/pdu.rb
```diff
@@ -194,13 +194,13 @@ def parse_ldap_result(sequence)
...
   def parse_extended_response(sequence)
-    sequence.length >= 3 or raise Net::LDAP::PDU::Error, "Invalid LDAP result length."
+    sequence.length.between?(3, 5) or raise Net::LDAP::PDU::Error, "Invalid LDAP result length."
     @ldap_result = {
       :resultCode => sequence[0],
       :matchedDN => sequence[1],
       :errorMessage => sequence[2],
     }
-    @extended_response = sequence[3]
+    @extended_response = sequence.last
```
HarlemSquirrel, line 203: What's the idea here?

HarlemSquirrel (commented):
Thanks for tackling this! 

Looks like tests are failing and I have a few questions

lib/net/ldap.rb
zeroSteiner: Without this change in place, the reply would cause an exception to be raised when it was being parsed.

lib/net/ldap/pdu.rb
zeroSteiner: There is an optional field that's included when the target is Active Directory that shifts the index. This accounts for both cases by accessing it as the last member.

zeroSteiner:
Let me add the scripts I used for testing. IIRC I tested both with OpenLDAP and Active Directory.

<details>
<summary>test_password_change.rb</summary>

```ruby
$LOAD_PATH.unshift 'lib'

require 'rubygems'
require 'net/ldap'
require 'pry'

openldap = {
     :host => '127.0.0.1',
     :port => 389,
     :auth => {
           :method => :simple,
           :username => "cn=admin,dc=example,dc=org",
           :password => "adminpassword"
     }
}


ldap = Net::LDAP.new openldap
ldap.bind
result = ldap.password_modify(
  dn: 'uid=johndoe,ou=users,dc=example,dc=org',
  #old_password: 'admin',
  new_password: 'Password1!'
)
puts "password_modify: #{result}"
```
</details>

<details>
<summary>test_whoami.rb</summary>

```ruby
cat test_whoami.rb 
$LOAD_PATH.unshift 'lib'

%Q{
podman run --rm --name openldap \
  -p 1389:1389 \
  --env LDAP_ADMIN_USERNAME=admin \
  --env LDAP_ADMIN_PASSWORD=adminpassword \
  --env LDAP_USERS=customuser \
  --env LDAP_PASSWORDS=custompassword \
  --env LDAP_ROOT=dc=example,dc=org \
  --env LDAP_ADMIN_DN=cn=admin,dc=example,dc=org \
  bitnami/openldap:2.5.17
}

require 'rubygems'
require 'net/ldap'
require 'pry'

openldap = {
     :host => '127.0.0.1',
     :port => 1389,
     :auth => {
           :method => :simple,
           :username => "cn=admin,dc=example,dc=org",
           :password => "adminpassword"
     }
}

openldap2 = {
     :host => '127.0.0.1',
     :port => 389,
     :auth => {
           :method => :simple,
           :username => "uid=johndoe,ou=users,dc=example,dc=org",
           :password => "password123"
     }
}

adds = {
     :host => '192.168.159.10',
     :port => 389,
     :auth => {
           :method => :simple,
           :username => "CN=Alice Liddle,CN=Users,DC=msflab,DC=local",
           :password => "Password1!"
     }
}

ldap = Net::LDAP.new openldap
ldap.bind
puts "[openldap]  whoami: #{ldap.whoami}"

#ldap = Net::LDAP.new openldap2
#ldap.bind
#puts "[openldap2] whoami: #{ldap.whoami}"

ldap = Net::LDAP.new adds
ldap.bind
puts "[adds]      whoami: #{ldap.whoami}"
```
</details>

lib/net/ldap.rb
HarlemSquirrel: I guess this is fine if it doesn't break other queries

lib/net/ldap/pdu.rb
HarlemSquirrel: OK, as long as it works with other LDAP directories

zeroSteiner:
@HarlemSquirrel tests are passing now! There was an issue with parsing the response from password-modify requests that deviated from the RFC which defines that it's in it's own format. The `#extended_response` attribute should be the unparsed data to closely adhere to the RFC spec, and extracting the password from it should require parsing it as per the other RFC spec.

HarlemSquirrel (approved):
Thanks for working on this!

PR #430, opened 2024-10-08
Add support for UTF-8 encoded passwords

frankwalentowski:
Fixing creation of SHA1/SHA256 digest in Net::LDAP::Password when given password string has an UTF8 encoding.


HarlemSquirrel (commented):
Thanks for this! 

The branch now needs to be updated for tests to run but it looks like you don't have your account set so I can update that branch for you for this PR. Please update this branch from the latest master here so the tests can run.

HarlemSquirrel:
@frankwalentowski thanks for working on this! Do you have a moment to update this branch from the latest master?

frankwalentowski:
Sorry, I've been distracted with other things for the last few months.

HarlemSquirrel:
@frankwalentowski looks like just one Rubocop offense and this should be good to go

PR #432, opened 2024-10-26
#431, Add `ostruct` as a dependency to the gemspec

### LINKED ISSUE #431: Request to Add ostruct as a Dependency in net-ldap
## Background:
With the release of Ruby 3.3.5, users are now receiving warnings when using libraries that utilize ostruct without explicitly declaring it as a dependency in their gemspec. This change highlights the need for better clarity on dependencies and may lead to confusion among users regarding the operational aspects of the net-ldap gem.

The current version of net-ldap (0.19.0) does not include ostruct in its gemspec file, which can cause warning messages when used in Ruby environments upgraded to version 3.3.5 or later.

## Proposed Change:
I recommend reaching out to the maintainers of net-ldap to request that they add ostruct as a dependency in the gemspec. This addition would help prevent warnings for users while ensuring clear communication regarding the required components of the library.

## Example Implementation:
In the net-ldap.gemspec file, adding the following line would suffice:

```
s.add_dependency("ostruct")
```

Ivanov-Anton:
## Description

This commit adds `ostruct` as an explicit dependency in the net-ldap gemspec.

With the release of Ruby 3.3.5 and later versions, users of net-ldap may encounter warnings related to the use of `ostruct` if it is not declared as a dependency. By including `ostruct`, we aim to enhance clarity regarding the gem's requirements and prevent any runtime issues related to missing dependencies.

## Related links

closes #431

### DIFF
--- a/net-ldap.gemspec
+++ b/net-ldap.gemspec
@@ -29,6 +29,7 @@ the most recent LDAP RFCs (4510-4519, plutions of 4520-4532).}
   s.required_ruby_version = ">= 2.0.0"
   s.summary = %q{Net::LDAP for Ruby (also called net-ldap) implements client access for the Lightweight Directory Access Protocol (LDAP), an IETF standard protocol for accessing distributed directory services}
 
+  s.add_dependency("ostruct")
   s.add_development_dependency("flexmock", "~> 1.3")
   s.add_development_dependency("rake", "~> 12.3.3")
   s.add_development_dependency("rubocop", "~> 1.48")


Ivanov-Anton:
in our software after updating ruby to 3.3.5 there is warning that require to add the ostruct gem as an explicit dependency

warning example:

```
warning: .../ostruct.rb was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add ostruct to your Gemfile or gemspec to silence this warning.
Also please contact the author of net-ldap-0.19.0 to request adding ostruct into its gemspec.
```

HarlemSquirrel (approved):
Thanks!

Ivanov-Anton:
@HarlemSquirrel Thanks, waiting a new release.

